PR #6, opened 2024-05-11
Added custom user model with customer user manager

bartczak-pa:
This overwrites default django behaviour and allows to create users based on email instead of username.

Can we check if this code is okay?


jacoor:
@bartczak-pa please see tests results. Those failed.
https://github.com/pymasterspl/reddit/actions/runs/9045129435/job/24854541143 
If you look here, it failed on migrations. This will be checked only after all is green. Please fix migrations.

bartczak-pa:
@jacoor Migrations are fixed and code is passing tests now.

jacoor (changes requested):
The code looks good.
However, Ruff is configured on purpose to enforce typings. 
Please, do fix ruff errors and avoid #noqa unless really necessary. 
I pushed an example of type updates for you, please fix other missing types. 
If you have issues with typing, you can try to use AI to fix 
- copilot - ask it to add typings for you (this works)
- or copy the code to Chat GPT and ask to add types.
Of course, do not forget to use `ruff check --fix` for other issues. 
Again, only use #noqa if this is really necessary, like:
`users/models.py:29:19: TRY003 Avoid specifying long messages outside the exception class`, however, for this one google has very good answer you can find after couple seconds search:
https://docs.astral.sh/ruff/rules/raise-vanilla-args/
(I put "how to fix TRY003 Avoid specifying long messages outside the exception class`" as a question to google).

users/models.py
```diff
@@ -1,3 +1,41 @@
...
+from django.contrib.auth.base_user import BaseUserManager
+from django.contrib.auth.models import AbstractUser
+from django.db import models
 
-# Create your models here.
+
+class UserManager(BaseUserManager):
+    use_in_migrations: bool = True
+
+    def _create_user(self: "UserManager", email: str, password: str, **extra_fields: int) -> "User":
```
jacoor, line 11: Here you have an example of added typings.

users/models.py
```diff
@@ -1,3 +1,41 @@
...
+        if not email:
+            message: str = "Users must have an email address"
```
jacoor, line 13: And fixed ruff errors, this one is EM101

jacoor:
@bartczak-pa you need to resolve conflicts. After that, tests will start to run and we will know if the code good. 
Let me know if you need help.

jacoor (approved):

jacoor:
Great work. Now merge away!

bartczak-pa:
Typings and conflicts are resolved.

PR #12, opened 2024-05-15
Created app: core

swietlikm:
- Created core models: Community, Post, PostVote, Tag
- added tests of tags and posts score
- added dependencies to Poetry


core/migrations/0001_initial.py
```diff
@@ -0,0 +1,156 @@
...
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [  # noqa: RUF012
+        ("contenttypes", "0002_remove_content_type_name"),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
+
+    operations = [  # noqa: RUF012
```
jacoor: MIgrations are excluded from ruff tests: https://github.com/pymasterspl/reddit/blob/494c9c9a743d0090c3eff5ebd50df9af91a2f682/pyproject.toml#L24
please remove those #noqa notes and if this still fails, we need to fix ruff config.

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class GenericModel(models.Model):
+    user = models.ForeignKey(User, on_delete=models.SET_NULL, null=True)
+    is_active = models.BooleanField(default=True)
+    is_locked = models.BooleanField(default=False)
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:  # noqa: D106
```
jacoor: Remove this and add to ruff config in pyproject.toml 
https://github.com/pymasterspl/reddit/blob/494c9c9a743d0090c3eff5ebd50df9af91a2f682/pyproject.toml#L32

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+        abstract = True
+
+
+class Community(GenericModel):
+    name = models.CharField(max_length=255)
+
+    def __str__(self) -> str:  # noqa: ANN101, D105
```
jacoor: Fix please. 
ANN101: https://docs.astral.sh/ruff/rules/missing-type-self/
D105 - add to ingores in ruff.

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class Tag(models.Model):
+    name = models.SlugField()
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    content_object = GenericForeignKey("content_type", "object_id")
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:  # noqa: D106
```
jacoor: as above

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+        indexes: ClassVar[list[tuple[str, ...]]] = [
+            models.Index(fields=["content_type", "object_id"]),
+        ]
+
+    def __str__(self) -> str:  # noqa: ANN101, D105
```
jacoor: as above

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+    title = models.CharField(max_length=255)
+    content = models.TextField()
+
+    tags = GenericRelation(Tag, related_query_name="posts")
+    image = models.ImageField(blank=True)
+    parent = GenericRelation("self", related_query_name="children")
+    version: str = models.CharField(max_length=32)
+
+    def __str__(self) -> str:  # noqa: ANN101, D105
```
jacoor: as above

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+        return f"@{self.user}: {self.title}"
+
+    def save(self, *args, **kwargs) -> None:  # noqa: ANN101, ANN002, ANN003
```
jacoor: Use google, fix. 
https://docs.astral.sh/ruff/rules/missing-type-args/

core/tests/test_models.py
```diff
@@ -0,0 +1,198 @@
...
+
+from core.models import Community, Post, PostVote, Tag
+
+User = get_user_model()
+
+UPVOTE_SCORE = 1
+DOWNVOTE_SCORE = -1
+MULTIPLE_VOTES_SCORE = 2
+SCORE_3 = 3
+MIXED_VOTES_SCORE = 0
```
jacoor: Please change those static names to be more meaningful.....
Ie SCORE_3 says what is score number. 
MIXED_VOTES_SCORE requires lookup so the test readability is bad.
Actually, are those needed? Seems it will be more readable to simply use numbers where needed, don't you think?

core/tests/test_models.py
```diff
@@ -0,0 +1,198 @@
...
+    assert post.score == MIXED_VOTES_SCORE
+
+
+@pytest.mark.django_db()
+def test_post_score_upvote(post: Post, user: User) -> None:
+    vote = PostVote.objects.create(type=PostVote.UPVOTE)
+    vote.user.add(user)
+    vote.post.add(post)
+    post.refresh_from_db()
+    assert post.score == UPVOTE_SCORE
```
jacoor: Please change to number, will be more readable. And all occurrences I think.

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Add new tags
+        new_tags = current_tags - existing_tags
+        for tag in new_tags:
+            Tag.objects.create(name=tag, content_object=self)
+
+        if self.created_at != self.updated_at:
+            self.version = self.generate_version()
+
+    def generate_version(self) -> str:  # noqa: ANN101
```
jacoor: as above

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+            object_id=self.id,
+        ).delete()
+
+        # Add new tags
+        new_tags = current_tags - existing_tags
+        for tag in new_tags:
+            Tag.objects.create(name=tag, content_object=self)
+
+        if self.created_at != self.updated_at:
+            self.version = self.generate_version()
```
jacoor: This is not saved. 
Moreover, it will create new version on every save click after first update, even if there is no change. 
Actually, I am thinking this condition is not needed, no? Worst case it will generate md5 again, but since no change in text, we loose bit compute power on hash generation. 
Either way, it needs save, so prob move before super.save()

core/tests/test_models.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+@pytest.mark.django_db()
+def test_post_score_initial(post: Post) -> None:
+    assert post.score == MIXED_VOTES_SCORE
+
+
+@pytest.mark.django_db()
+def test_post_score_upvote(post: Post, user: User) -> None:
+    vote = PostVote.objects.create(type=PostVote.UPVOTE)
```
jacoor: Would be good to have a method on Post like "vote", so we don't have to repeat this code.

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+    def generate_version(self) -> str:  # noqa: ANN101
+        data = f"{self.title}{self.content}"
+        return hashlib.sha256(data.encode()).hexdigest()
+
+    @property
+    def up_votes(self) -> int:  # noqa: ANN101
+        return self.post_votes.filter(type=PostVote.UPVOTE).count()
+
+    @property
+    def down_votes(self) -> int:  # noqa: ANN101
```
jacoor: make those number fields please. Do count when vote is added. This way we won't kill DB. 
Imagine situation when we display a post with 100 comments on page, and every comment/post needs to run this count. 
Score can be dynamic, though.

core/tests/test_models.py
```diff
@@ -0,0 +1,198 @@
...
+    assert post.score == UPVOTE_SCORE
+
+
+@pytest.mark.django_db()
+def test_post_score_downvote(post: Post, user: User) -> None:
+    vote = PostVote.objects.create(type=PostVote.DOWNVOTE)
+    vote.user.add(user)
+    vote.post.add(post)
+    post.refresh_from_db()
+    assert post.score == DOWNVOTE_SCORE
```
jacoor: same as above, number please.

core/tests/test_models.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+@pytest.mark.django_db()
+def test_post_score_multiple_votes(post: Post, user: User) -> None:
+    another_user = User.objects.create_user(
+        email="anotheruser@test.pl",
+        password=generate_random_password(),
+    )
+
+    vote1 = PostVote.objects.create(type=PostVote.UPVOTE)
```
jacoor: Yup, we need a method on post to add vote.

core/tests/test_models.py
```diff
@@ -0,0 +1,198 @@
...
+    vote1.user.add(user)
+    vote1.post.add(post)
+
+    vote2 = PostVote.objects.create(type=PostVote.UPVOTE)
+    vote2.user.add(another_user)
+    vote2.post.add(post)
+
+    post.refresh_from_db()
+    assert post.score == MULTIPLE_VOTES_SCORE
```
jacoor: Number, please.

core/models.py
```diff
@@ -0,0 +1,125 @@
...
+        on_delete=models.CASCADE,
+        related_name="posts",
+    )
+    title = models.CharField(max_length=255)
+    content = models.TextField()
+
+    tags = GenericRelation(Tag, related_query_name="posts")
+    image = models.ImageField(blank=True)
+    parent = GenericRelation("self", related_query_name="children")
+    version: str = models.CharField(max_length=32)
```
jacoor: Please add help_text and make it read only.

jacoor (changes requested):
Nice work, I really like the tests. 
Required: please fix types. 
Optional: the optimization, like storing votes count on post and adding vote method, can be done in another task if you want, or with this. 
If you want to do it in separate task, please create one for yourself in clickup and link here for review before this will be approved for merge.

jacoor:
One more thing. 
Please make commits often, and per functionality. 
IE. 
- added post post model
- added vote model
- added tests for
- fixed tests
- update poetry

This way we see what is changing it the project, we can revert a single thing. One huge commit with description "initial" is really not saying too much. 

I know, we can squash migrations, but still, it's a good practice to make commits often. When you play a game, you do make saves often, don't you? :-)

core/models.py
```diff
@@ -116,10 +149,23 @@ class PostVote(models.Model):
...
+    class Meta:
+        unique_together: ClassVar[list[str]] = ["post", "user"]
+
+    def __str__(self: "PostVote") -> str:
+        return f"@{self.user}: {self.choice} for post: {self.post}"
+
+    def save(self: "PostVote", *args: int, **kwargs: int) -> None:
+        super().save(*args, **kwargs)
+        post_vote = PostVote.objects.filter(post=self.post)
+        self.post.up_votes = post_vote.filter(choice=PostVote.UPVOTE).count()
```
jacoor: This should be optimized a bit, using "update" method.
1. since vote is either up or down, it could count only this no?
2. instead of running save -> which will actually alter updated at, we need to run update... using this:
https://docs.djangoproject.com/en/5.0/topics/db/queries/#updating-multiple-objects-at-once
```
post.objects.filter(pk=self.post.pk).update(down_votes={some number}
```
This has the following advantages:
- does not call any signals on objects
- does not call save() on objects
- does not update changed_at value (which will get changed when save is called) and which we need to avoid in this scenario
- it simply alters “silently” one column on the object in db

pyproject.toml
```diff
@@ -22,15 +23,15 @@ requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.ruff]
-exclude = ["scripts/", "manage.py", "asgi.py", "wsgi.py", "migrations/"]
+exclude = ["scripts/", "manage.py", "asgi.py", "wsgi.py", "*/migrations/"]
```
jacoor, line 26: +1

pyproject.toml
```diff
@@ -47,4 +48,5 @@ line-ending = "auto"
     "S101", # asserts allowed in tests...
     "ARG", # Unused function args -> fixtures nevertheless are functionally relevant...
     "FBT", # Don't care about booleans as positional arguments in tests, e.g. via @pytest.mark.parametrize()
+    "PLR2004", # Ruff is requesting to replace asserted integers for a constants
```
jacoor, line 51: +1

jacoor (changes requested):

core/models.py
```diff
@@ -0,0 +1,171 @@
...
+class Post(GenericModel):
+    community = models.ForeignKey(
+        Community,
+        on_delete=models.CASCADE,
+        related_name="posts",
+    )
+    title = models.CharField(max_length=255)
+    content = models.TextField()
+    tags = GenericRelation(Tag, related_query_name="posts")
+    image = models.ManyToManyField(Image, blank=True, related_name="posts")
```
jacoor: Hmm I think. I would not make it many too many. 
Post can have multipole images, but image can have only one post so it should be foreign key.
On the image itself, we may need to have "ordering" field, int, 99999999. This way image can have ordering if needed and newly added will be always last.

core/models.py
```diff
@@ -162,10 +138,27 @@ def __str__(self: "PostVote") -> str:
...
+        down_votes = post_votes.filter(choice=PostVote.DOWNVOTE).count()
+        Post.objects.filter(pk=self.post.pk).update(up_votes=up_votes, down_votes=down_votes)
+
+
+class Image(models.Model):
+    post = models.ForeignKey(Post, on_delete=models.CASCADE, related_name="images")
+    image = models.ImageField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    order = models.PositiveIntegerField(default=0)
```
jacoor: default 9999999 was defined on purpose. Now, every image will have zero. Imagine you have 5 images, sorted, 0 to 4. Now you add one more. And it has 0... where it will go?
if it has 9999999 it will be last.

jacoor (changes requested):

core/models.py
swietlikm: Understood and agreed. Thank you.

jacoor (approved):

PR #15, opened 2024-05-25
Feature/users login

knop-k:
- added user control from the navbar
- added simple style to errors for login template


templates/base.html
```diff
@@ -23,8 +24,27 @@
                         aria-controls="navbarNav" aria-expanded="false" aria-label="Toggle navigation">
                     <span class="navbar-toggler-icon"></span>
                 </button>
-                {%  block nav %}
-                {%  endblock nav %}
+                {% block nav %}
+                <div class="collapse navbar-collapse" id="navbarNav">
+                    <ul class="navbar-nav ms-auto">
+                        {% if request.user.is_authenticated %}
+                            <li class="nav-item">
```
jacoor, line 31: please remove invalid links. Home page is home page, not logout or whatever. 
instead of / use # (so it's not an incorrect link) and change text to clearly state it has to be implemented. 
Leaving it as is will lead to us missing this spot while going further. In real life this means bug in production.

jacoor (changes requested):

templates/base.html
knop-k: good idea, thanks

jacoor (approved):

jacoor:
@knop-k merge when ready. I hit the same issue with ruff, so if I merge my first, you will need to fix conflict. :P

PR #16, opened 2024-05-26
allow running tests and ruff checks on draft pull requests

jacoor:


knop-k (approved):
good idea with these automatic draft tests

PR #17, opened 2024-05-26
Feature/user registration

swietlikm:
- created Registrtaion View
- created Registration Template
- added tests

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -4,10 +4,16 @@ from collections.abc import Generator
 
 import pytest
 from django.contrib.auth import get_user_model
+from django.test import Client
 
 from core.models import Community, Post
 
 User = get_user_model()
+
+
+@pytest.fixture()
+def client() -> type[Client]:
+    return Client()
 
 
 def generate_random_password(length: int = 12) -> str:
@@ -19,6 +25,11 @@ def generate_random_password(length: int = 12) -> str:
 @pytest.fixture()
 def generated_password() -> str:
     return generate_random_password()
+
+
+@pytest.fixture()
+def user_model() -> type[get_user_model()]:
+    return get_user_model()
 
 
 @pytest.fixture()
--- a/conftest.py
+++ b/conftest.py
@@ -4,16 +4,10 @@ from collections.abc import Generator
 
 import pytest
 from django.contrib.auth import get_user_model
-from django.test import Client
 
 from core.models import Community, Post
 
 User = get_user_model()
-
-
-@pytest.fixture()
-def client() -> type[Client]:
-    return Client()
 
 
 def generate_random_password(length: int = 12) -> str:
--- a/users/templates/users/registration.html
+++ b/users/templates/users/registration.html
@@ -9,7 +9,15 @@
             <div class="mx-auto col-10 col-md-8 col-lg-6 vh-100 mt-lg-3">
                 <form method="POST" action=".">
                     {% csrf_token %}
+
+                    {% if form.non_field_errors %}
+                        <div class="alert alert-danger" role="alert">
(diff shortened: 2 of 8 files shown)


conftest.py
```diff
@@ -4,12 +4,18 @@
...
 from django.contrib.auth import get_user_model
+from django.test import Client
 
 from core.models import Community, Post
 
 User = get_user_model()
 
 
+@pytest.fixture()
+def client() -> type[Client]:
```
jacoor: Why? 
we have this: https://pytest-django.readthedocs.io/en/latest/helpers.html#client-django-test-client

users/templates/users/registration.html
```diff
@@ -0,0 +1,21 @@
...
+{% load static %}
+
+{% block content %}
+
+    <div class="container">
+        <h3 class="mb-4">Registration</h3>
+        <div class="d-flex justify-content-center">
+            <div class="mx-auto col-10 col-md-8 col-lg-6 vh-100 mt-lg-3">
+                <form method="POST" action=".">
+                    {% csrf_token %}
```
jacoor, line 11: Missing non_field_errors
https://docs.djangoproject.com/en/5.0/ref/forms/api/#django.forms.Form.non_field_errors
https://stackoverflow.com/a/16215790 - it's and old one but still valid. It's a good practice to add non field error handling.

users/tests/test_registration.py
```diff
@@ -0,0 +1,63 @@
...
+) -> None:
+    data: dict = {
+        "email": "testuser@example.com",
+        "password1": "Pass2712!",
+        "password2": "Pass2712!",
+    }
+    response = client.post(register_url, data)
+    assert response.status_code == 302
+    assert response.url == reverse("home-page")
+    assert user_model.objects.filter(email="testuser@example.com").exists()
```
jacoor, line 37: Can you prove that this user did not exist earlier?

users/tests/test_registration.py
```diff
@@ -0,0 +1,63 @@
...
+    data: dict = {
+        "email": "",
+        "password1": "",
+        "password2": "",
+    }
+    response = client.post(register_url, data)
+    assert response.status_code == 200
+    assert "users/registration.html" in [t.name for t in response.templates]
+    assert not user_model.objects.filter(email="").exists()
+    assert "This field is required." in response.content.decode()
```
jacoor: Which field? Or how many error messages are there? Either way is good. 
Right now this test only show that one of the fields is required, doesn't show which one.

users/tests/test_registration.py
```diff
@@ -0,0 +1,63 @@
...
+    data: dict = {
+        "email": "testuser@example.com",
+        "password1": "TestPassword123",
+        "password2": "DifferentPassword123",
+    }
+    response = client.post(register_url, data)
+    assert response.status_code == 200
+    assert "users/registration.html" in [t.name for t in response.templates]
+    assert not user_model.objects.filter(email="testuser@example.com").exists()
+    assert "The two password fields didn’t match." in response.content.decode()
```
jacoor: hmm I would give up on testing passwords, as this is done in UserCreationForm.
But if you really want robust testing, also add check for invalid emain. 

And missing tests for already existing user.

jacoor (changes requested):

conftest.py
swietlikm: I do not understand, what you linked is reffering to An instance of a django.test.Client and the fixture is exactly this

jacoor: I mean, pytest-django, the package we have already, is providing this. We don't need second one in this fixture

jacoor (approved):

PR #21, opened 2024-05-28
feature/users-logout

swietlikm:
- logout confirmation template
- logout view
- added logout url to base template

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -32,7 +32,7 @@
                                 <a class="nav-link" href="#">User profile (to be implemented)</a>
                             </li>
                             <li class="nav-item">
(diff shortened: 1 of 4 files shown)


jacoor:
@swietlikm can you point me to task in click up for this? As far as I see, logout is assigned to @Pawelooo. 
So what happened here?

swietlikm:
@jacoor my error, anyway i spoke with @Pawelooo and he is fine with this

jacoor:
Cool, glad you two talked. 
It would be good to add a test here. Your call. Approved.

jacoor (approved):

PR #22, opened 2024-05-30
Feature/community members

oskarolaszczyk:
Task ID: 8694mw7jg
Model, tests - without different privileges depending on type


core/models.py
```diff
@@ -144,7 +144,10 @@ def save(self: "PostVote", *args: int, **kwargs: int) -> None:
         post_votes = PostVote.objects.filter(post=self.post)
         up_votes = post_votes.filter(choice=PostVote.UPVOTE).count()
         down_votes = post_votes.filter(choice=PostVote.DOWNVOTE).count()
-        Post.objects.filter(pk=self.post.pk).update(up_votes=up_votes, down_votes=down_votes)
```
jacoor, line 147: Please revert this format change. It's just whitespacing change which  so should not be committed. You may need to fine tune your editor settings.

core/models.py
```diff
@@ -165,3 +168,30 @@ def image_url(self: "Image") -> str:
             return self.image.url
         except ValueError:
             return ""
+
+
+class CommunityMember(models.Model):
+    ADMIN: ClassVar[str] = "admin"
+    MODERATOR: ClassVar[str] = "moderator"
+    MEMBER: ClassVar[str] = "member"
```
jacoor:
```suggestion
    MEMBER: ClassVar[str] = "100_member"
    MODERATOR: ClassVar[str] = "200_moderator"
    ADMIN: ClassVar[str] = "300_admin"
```
this way we can use number comparison if user has certain privileges. 
```
a = "100_x"
int(a.split("_")[0]) > 1
```
Also we are keeping member smallest as it has least privileges.

core/models.py
```diff
@@ -165,3 +168,30 @@ def image_url(self: "Image") -> str:
...
+        (MODERATOR, "Moderator"),
+        (MEMBER, "Member"),
+    ]
+
+    community = models.ForeignKey(
+        Community,
+        on_delete=models.CASCADE,
+        related_name="memberships",
+    )
+    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="communities")
```
jacoor: Are you sure about using on_delete cascade here? 
And on community?

core/models.py
```diff
@@ -165,3 +168,30 @@ def image_url(self: "Image") -> str:
             return self.image.url
         except ValueError:
             return ""
+
+
+class CommunityMember(models.Model):
```
jacoor, line 172: Should it not be used in Community as users M2M with through model?

core/tests/test_models.py
```diff
@@ -131,3 +132,42 @@ def test_tags_removed_on_post_update(user: User, community: Community) -> None:
...
+
+@pytest.mark.django_db()
+def test_create_community_user(user: User, community: Community) -> None:
+    community_user: CommunityMember = CommunityMember.objects.create(
+        user=user,
+        community=community,
+        role=CommunityMember.ADMIN,
+    )
+
+    assert community_user.user == user
```
jacoor: Hmm
this test would be much better if asserts were from community not from community_user (also mind name).
Actually this test does not make sense as it only checks if entry in db was created, and not even refreshing it. 
So the proper way would be (pseudocode)
assert community.users.filter(user = user) == user

jacoor (changes requested):

core/models.py
jacoor: This will help: https://akademiait.com.pl/zrozumiec-django-relacje-pomiedzy-modelami-wiele-do-wielu-manytomany-porady-i-triki/

core/models.py
oskarolaszczyk: I think so - if we delete a user, how can he be a member of the community later, just as we delete the community, the memberships should also be deleted

jacoor: True, thanks, I got confused.

core/models.py
oskarolaszczyk: Should I drop the related manager or using both?

oskarolaszczyk: ok, nvm

jacoor (approved):

PR #28, opened 2024-06-05
Feature/online status

swietlikm:
- Enchanced user model by last_activity field
- added property "is_online"
- added property "last_activity_ago" to display when user was active last time
- added LAST_ACTIVITY_ONLINE_LIMIT_MINUTES = 15 to settings.py
- added Middleware to automatically update last_activity of a user


pyproject.toml
```diff
@@ -50,6 +50,5 @@ line-ending = "auto"
     "S101", # asserts allowed in tests...
     "ARG", # Unused function args -> fixtures nevertheless are functionally relevant...
     "FBT", # Don't care about booleans as positional arguments in tests, e.g. via @pytest.mark.parametrize()
-    "PLR2004", # Ruff is requesting to replace asserted integers for a constants
```
jacoor, line 53: why removed?

reddit/settings.py
```diff
@@ -148,3 +149,6 @@
 # http://django-crispy-forms.readthedocs.io/en/latest/install.html#template-packs
 CRISPY_TEMPLATE_PACK = "bootstrap5"
 CRISPY_ALLOWED_TEMPLATE_PACKS = "bootstrap5"
+
+# Set time after last  activity on page, when user change status into offline (minutes)
+LAST_ACTIVITY_ONLINE_LIMIT = 15
```
jacoor: what is the unit of it?  15 of what? Hours, days, seconds? You can use string here which in turn will fix the ruff problem, PLR2004. Do not use comment to declare unit, it should be in the constant. Or do it another way, either way, this constant and it value should carry whole information, without the need to look up the comment.

users/admin.py
```diff
@@ -1,14 +1,14 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin as DjangoUserAdmin
 
-from users.models import User
+from .models import User
```
jacoor: why this change? Please revert.

users/models.py
```diff
@@ -44,3 +47,26 @@ class User(AbstractUser):
     objects = UserManager()
     username: None = None
     email: str = models.EmailField(unique=True)
+    last_activity = models.DateTimeField(default=timezone.now)
```
jacoor: remove default, use auto_now_add.

users/models.py
```diff
@@ -44,3 +47,26 @@ class User(AbstractUser):
...
+
+    def update_last_activity(self: "User") -> None:
+        self.last_activity = timezone.now()
+        self.save(update_fields=["last_activity"])
+
+    @property
+    def is_online(self: "User") -> bool:
+        online_limit = timezone.now() - timedelta(minutes=settings.LAST_ACTIVITY_ONLINE_LIMIT)
```
jacoor: Here is a problem. To really see what is LAST_ACTIVITY_ONLINE_LIMIT I need to go from settings to here. I don't trust the comment in settings. Please either add unit to settings or suggest other way of fixing this issue.

pyproject.toml
```diff
@@ -33,7 +33,7 @@ target-version = "py311"
 
 [tool.ruff.lint]
 select = ["ALL"]
-ignore = ["D104", "D100", "D103", "D211", "D213", "D101", "D102", "D105", "D106"]
+ignore = ["D104", "D100", "D103", "D211", "D213", "D101", "D102", "D105", "D106", "D107", "PLR2004"]
```
jacoor: Do not add PLR2004. Add it inline in the file, probably settings.py

users/models.py
```diff
@@ -44,3 +47,26 @@ class User(AbstractUser):
     objects = UserManager()
     username: None = None
     email: str = models.EmailField(unique=True)
+    last_activity = models.DateTimeField(default=timezone.now)
+
+    def update_last_activity(self: "User") -> None:
+        self.last_activity = timezone.now()
+        self.save(update_fields=["last_activity"])
```
jacoor: I would use update here not save I'd say, to avoid post save hooks.

jacoor (changes requested):

pyproject.toml
```diff
@@ -14,6 +14,7 @@ pytest-django = "^4.8.0"
 pillow = "^10.3.0"
 django-crispy-forms = "^2.1"
 crispy-bootstrap5 = "^2024.2"
+freezegun = "^1.5.1"
```
jacoor: Please remove, since it is not used.

jacoor (changes requested):

reddit/settings.py
```diff
@@ -148,3 +149,6 @@
 # http://django-crispy-forms.readthedocs.io/en/latest/install.html#template-packs
 CRISPY_TEMPLATE_PACK = "bootstrap5"
 CRISPY_ALLOWED_TEMPLATE_PACKS = "bootstrap5"
+
+# Set time after last  activity on page, when user change status into offline (minutes)
```
jacoor: remove the comment, it does not bring any value now.

jacoor (approved):

PR #31, opened 2024-06-07
Feature/user saved posts

swietlikm:
Added new model: SavedPost 
Added new methods for Post: save_post and remove_saved_post
Added method for User model: get_saved_posts
Added new model tests


core/models.py
```diff
@@ -120,6 +120,13 @@ def vote(self: "Post", user: User, choice: str) -> None:
     def get_images(self: "Post") -> models.QuerySet:
         return Image.objects.filter(post=self)
 
+    def save_post(self: "Post", user: User) -> "SavedPost":
```
jacoor: Architecture wise it would be better to have this method on SavedPost model. This way post model does not know anything about SavedPost, which is good cause we can easily remove savedPost or copy post model to new project. 
So please move this method into SavedPost

core/models.py
```diff
@@ -120,6 +120,13 @@ def vote(self: "Post", user: User, choice: str) -> None:
     def get_images(self: "Post") -> models.QuerySet:
         return Image.objects.filter(post=self)
 
+    def save_post(self: "Post", user: User) -> "SavedPost":
+        saved_post, created = SavedPost.objects.get_or_create(user=user, post=self)
+        return saved_post
+
+    def remove_saved_post(self: "Post", user: User) -> None:
```
jacoor: Architecture wise it would be better to have this method on SavedPost model. This way post model does not know anything about SavedPost, which is good cause we can easily remove savedPost or copy post model to new project. 
So please move this method into SavedPost

users/models.py
```diff
@@ -44,3 +44,6 @@ class User(AbstractUser):
     objects = UserManager()
     username: None = None
     email: str = models.EmailField(unique=True)
+
+    def get_saved_posts(self: "User") -> models.QuerySet:
```
jacoor: Architecture wise, it would be better to have this method on SavedPost model. This way user model does not know anything about SavedPost, which is good because we can easily remove savedPost or copy user model to new project. 
So please move this method into SavedPost

jacoor (changes requested):

users/models.py
jacoor: You missed this one.

jacoor (changes requested):

jacoor (approved):

PR #35, opened 2024-06-13
Add username field to User model

yanazPL:
I' have:
  1. added username field to User model.
  2. updated registration/login/logout/reset password views.
  3. updated tests
  4. created migrations

When it comes to migrations:
  - 0005 migration adds `username(null=True)` field
  - 0006 migration populates User models with usernames infered from emails
  - 0007 makes username unique
  

Thanks for
https://stackoverflow.com/questions/56934804/best-practice-when-add-a-new-unique-field-to-an-existing-django-model/56935415#56935415)


users/tests/test_login.py
```diff
@@ -0,0 +1,28 @@
...
+    login_url = reverse("login")
+    response = client.get(login_url)
+    assert "users/login.html" in [t.name for t in response.templates]
+
+
+@pytest.mark.django_db()
+def test_login_view(client: Client, user: User) -> None:
+    login_url = reverse("login")
+    data = {
+        "username": user.username,
```
jacoor, line 21: Please change. We use email as login.

users/views.py
```diff
@@ -22,13 +20,6 @@ class LoginUserView(LoginView):
     def get_success_url(self: "LoginUserView") -> str:
         return reverse_lazy("home-page")
 
-    def form_valid(self: "LoginUserView", form: Form) -> HttpResponse:
-        valid = super().form_valid(form)
-        email, password = form.cleaned_data.get("email"), form.cleaned_data.get("password1")
```
jacoor, line 27: @yanazPL your task was to add username, not to change login functionality. Revert.

users/views.py
```diff
@@ -38,8 +29,9 @@ class UserRegistrationView(FormView):
 
     def form_valid(self: "UserRegistrationView", form: UserRegistrationForm) -> HttpResponse:
         form.save()
+        username = form.cleaned_data.get("username")
         email = form.cleaned_data.get("email")
-        messages.success(self.request, f"Account created for {email}!")
+        messages.success(self.request, f"Account created for {username} with: {email}!")
```
jacoor, line 34: As above

jacoor (changes requested):

PR #39, opened 2024-06-15
Task 8694r3zk8 - Post-Display-Counter plus technical views

tgrycuk:
This pull request introduces a feature to count and record the number of views each post receives, along with technical views.


core/views.py
```diff
@@ -1 +1,39 @@
...
+
+    template_name = "post-detail.html"
+
+    def get_context_data(self: "PostDetailView", **kwargs: dict) -> dict:
+        context = super().get_context_data(**kwargs)
+
+        post_id = self.kwargs.get("pk")
+        post = get_object_or_404(Post, pk=post_id)
+        post.view_count += 1
+        post.save(update_fields=["view_count"], skip_version_check=True)
```
jacoor: I am thinking it should be update as here: 
https://github.com/pymasterspl/reddit/blob/fe257582955c20ea67e95b781a348c43d0d6785e/users/models.py#L53-L54
Reasoning: whenever save is run, all signals are run - post saves and all this that django runs. We don't need that, we only want to bump +1 to the read count. 
So please change to update.

jacoor: This should be a method on post itself, not in the view. The method itself is ok, but really, should be in model - after all it's logic on the model. 
(lines 31 to 34)

Then, the call to post.update_display_counter() should be most likely called in view get_object method, so tou will not even need get_context_data.

core/models.py
```diff
@@ -78,12 +78,14 @@ class Post(GenericModel):
         max_length=32,
         help_text="Hash of the title + content to prevent overwriting already saved post",
     )
+    view_count = models.IntegerField(default=0)
 
     def __str__(self: "Post") -> str:
         return f"@{self.author}: {self.title}"
 
     def save(self: "Post", *args: int, **kwargs: int) -> None:
-        if self.pk is not None and self.generate_version() == self.version:
```
jacoor: This change won't be needed, please check below.

jacoor (changes requested):

core/views.py
swietlikm: You can try to use F() expression that is great for db performance
[https://docs.djangoproject.com/en/5.0/ref/models/expressions/#f-expressions](url)

core/models.py
jacoor: Please revert the changes to save. Not needed now.

core/models.py
```diff
@@ -127,6 +131,9 @@ def vote(self: "Post", user: User, choice: str) -> None:
     def get_images(self: "Post") -> models.QuerySet:
         return Image.objects.filter(post=self)
 
+    def update_display_counter(self: "Post") -> None:
+        Post.objects.filter(pk=self.pk).update(display_counter=F("display_counter") + 1)
```
jacoor, line 133: Nice use of F.

jacoor (changes requested):

jacoor (approved):

PR #41, opened 2024-06-15
Add nickname field to User model

yanazPL:
I' have:

1. added nickname field to User model and updated UserManager
2. fixed LoginUserVIew
3. updated tests and forms
4. created migrations

When it comes to migrations:

- 0006 migration adds `nickname(null=True)` field
- 0007 migration populates User models with nicknames infered from emails
- 0008 makes nickname unique

Thanks for  
[https://stackoverflow.com/questions/56934804/best-practice-when-add-a-new-unique-field-to-an-existing-django-model/56935415#56935415](https://stackoverflow.com/questions/56934804/best-practice-when-add-a-new-unique-field-to-an-existing-django-model/56935415#56935415))


users/models.py
```diff
@@ -4,29 +4,35 @@
...
+    def _create_user(self: "UserManager", email: str, nickname:str, password: str, **extra_fields: int) -> "User":
         if not email:
             message: str = "Users must have an email address"
             raise ValueError(message)
+        if not nickname:
+            message: str = "Users must have a nickname"
+            raise ValueError(message)
         email = self.normalize_email(email)
-        user: User = self.model(email=email, **extra_fields)
+        user: User = self.model(email=email, nickname=nickname, password=password, **extra_fields)
```
jacoor: user: User = self.model(email=email, nickname=nickname, **extra_fields)

users/models.py
```diff
@@ -38,12 +44,28 @@ def create_superuser(self: "UserManager", email: str, password: str, **extra_fie
...
+    nickname = models.CharField(
+        max_length=150,
+        null=False,
+        blank=False,
+        unique=True,
+        help_text=_(
+            "Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only.",
+        ),
+        validators=[nickname_validator],
+        error_messages={
```
jacoor: This is not neeed, automatically provided by Django when unique = True

jacoor (changes requested):

jacoor (approved):

PR #44, opened 2024-06-16
 Ruff formatting and configuration

yanazPL:
Exclude D203 rule from `[tool.ruff.lint]` so there is no conflict with D205.

Then format whole codebase.

Be cutious about possible `ruff check path` / `ruff path` versus `ruff format path` conflicts.

This is how to use formatter:
```bash
ruff format                   # Format all files in the current directory.
ruff format path/to/code/     # Format all files in `path/to/code` (and any subdirectories).
ruff format path/to/file.py   # Format a single file.
```
More at:
https://docs.astral.sh/ruff/formatter/


jacoor (changes requested):
Lovely, thank you. 
1. Please update readme with the instructions on how ot use this formatter
2. Provide information how can we use it in CI/CD - how to run it in a way that it will not do changes, but will verify compliance and return error if issues found.

jacoor:
You are also welcome to change GH action to support #2 from above (this file: https://github.com/pymasterspl/reddit/blob/dev/.github/workflows/pep.yml)

jacoor:
@yanazPL I can handle GH action, but I need documentation in readme.

yanazPL:
> @yanazPL I can handle GH action, but I need documentation in readme.

I have tried to add formatting to GH action (pep.yml) - not sure if it's correct. I wanted to verify that with nektos/act project, but had some problems (https://github.com/nektos/act)

Docs updated.

Gotta solve conflicts.

pyproject.toml
```diff
@@ -33,7 +33,7 @@ target-version = "py311"
 
 [tool.ruff.lint]
 select = ["ALL"]
-ignore = ["D104", "D100", "D103", "D211", "D213", "D101", "D102", "D105", "D106", "D107"]
+ignore = ["D104", "D100", "D103", "D211", "D213", "D101", "D102", "D105", "D106", "D107", "D203", "ISC001", "COM812"]
```
jacoor, line 37: Can you explain those new exceptions, please? Here, in comment? Thanks!

jacoor (commented):
The GH action changes looks good, I will also give it a run. Seems they pass anyway so nice.

README.md
```diff
@@ -117,6 +117,29 @@ python manage.py loaddata {fixture_name}.json
...
+ruff check path/to/code/    # Lint all files in `path/to/code` (and any subdirectories).
+```
+
+
+This is how to use formatter:
+```bash
+ruff format                   # Format all files in the current directory.
+ruff format path/to/code/     # Format all files in `path/to/code` (and any subdirectories).
+ruff format path/to/file.py   # Format a single file.
+ruff format --check           # Reports if there are differences between current and in future formatted files.
```
jacoor:
```suggestion
ruff format --check           # Checks if formatting is proper and reports violations. Does not change files.
```

pyproject.toml
yanazPL: > Can you explain those new exceptions, please? Here, in comment? Thanks!

These linter rules are ignored because they can get in conflict with formatter. They don't seem crucial.

I'm also ignoring them to supress `ruff format` warnings.

Exact rules:
https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/
https://docs.astral.sh/ruff/rules/prohibited-trailing-comma/

Ruff docs on linter vs formatter conflicts:
https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules

jacoor (approved):

PR #45, opened 2024-06-16
Feature/post details

swietlikm:
Created post.html with standalone code of post instance
Modified post-detail to include post.html
Modified Post model to count all comments for single post
Modified base.html so the static styles are loaded after bootstrap to override custom styles


core/models.py
```diff
@@ -132,6 +132,16 @@ def get_images(self: "Post") -> models.QuerySet:
...
+    def children_count(self: "Post") -> int:
+        def count_descendants(post: "Post") -> int:
+            children = post.children.all()
+            total_children = children.count()
+            for child in children:
+                total_children += count_descendants(child)
+            return total_children
+
+        return count_descendants(self)
+
```
jacoor, line 145: Hmm this has potential to kill the DB.
I wonder if we should not store children amount as int on parent itself, this way we don't run count every time...
That's for future refactor. 
Nice use of recurency btw. 
```suggestion
    @property
    def children_count(self: "Post") -> int:
        def count_descendants(post: "Post") -> int:
            children = post.children.all()
            total_children = children.count()
            for child in children:
                total_children += count_descendants(child)
            return total_children

        return count_descendants(self)


```

jacoor: I'd only add this @property

jacoor (changes requested):
small change on this @proprty
We do have a DB expert in community, we need to ask him about optimizing this count.

jacoor (approved):

PR #48, opened 2024-06-18
added Post Saving feature for logged users

swietlikm:
- created PostSaveView
- crated is_saved_by filter
- added is_saved_by method to Post model
- modified template of post-header.html with save/unsave feature

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -143,7 +143,7 @@ class Post(GenericModel):
 
         return count_descendants(self)
 
-    def is_saved_by(self: "Post", user: User) -> bool:
+    def is_saved(self: "Post", user: User) -> bool:
         return SavedPost.objects.filter(user=user, post=self).exists()
 
 
--- a/core/models.py
+++ b/core/models.py
@@ -143,6 +143,9 @@ class Post(GenericModel):
 
         return count_descendants(self)
 
+    def is_saved_by(self: "Post", user: User) -> bool:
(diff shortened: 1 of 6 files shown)


core/models.py
```diff
@@ -143,6 +143,9 @@ def count_descendants(post: "Post") -> int:
 
         return count_descendants(self)
 
+    def is_saved_by(self: "Post", user: User) -> bool:
+        return SavedPost.objects.filter(user=user, post=self).exists()
```
jacoor, line 147: What does this do? It is a misleading name, I would expect it to return the name who saved the post, not boolean. 
Please change naming.

core/templatetags/is_saved_by.py
```diff
@@ -0,0 +1,15 @@
...
+
+from core.models import Post
+
+register = template.Library()
+
+User = get_user_model()
+
+
+@register.filter
+def is_saved_by(post: Post, user: User) -> bool:
```
jacoor: Similar here, this name is misleading..
Maybe only "is_saved"? or is_post_saved?

jacoor (changes requested):

core/templatetags/is_saved_by.py
jacoor: Naming is one of the most difficult things. Thanks

jacoor (approved):

PR #49, opened 2024-06-18
Feature/comments

swietlikm:
- Refactored PostDetailsView with post method (to handle receiving comments) and get_context_data to display comments to posts
- added get_comments method to Post model
- created CommentForm in freshly created core/forms.py
- created comment.html and comment_tree.html to handle comments
- refactored post-detail.html to handle comments


core/models.py
```diff
@@ -143,6 +143,9 @@ def count_descendants(post: "Post") -> int:
 
         return count_descendants(self)
 
+    def get_comments(self: "Post") -> models.QuerySet:
+        return self.children.all()
```
jacoor, line 160: Missing is_active filtering. 
Actually, I am thinking we may need manager on GenericModel so we don't throw is active all over the place. 
- set default manager to filter is_active = Tue
- in admin use second manager, `objects_all` or similar name. 
https://docs.djangoproject.com/en/5.0/topics/db/managers/#modifying-a-manager-s-initial-queryset

Thoughts?

core/views.py
```diff
@@ -25,6 +28,38 @@ def get_object(self: "Post", queryset: QuerySet[Post] | None = None) -> Post:
...
+            new_comment.save()
+            return redirect("post-detail", pk=pk)
+
+        comments = post.get_comments()
+        context = {
+            "post": post,
+            "comments": comments,
+            "form": form,
+        }
+        return render(request, "post_detail.html", context)
```
jacoor: Hmm I know we can use render but...
Cold you use
```
html_content = render_to_string("post_detail.html", context)
return HttpResponse(html_content)
```
instead?

core/views.py
```diff
@@ -25,6 +28,38 @@ def get_object(self: "Post", queryset: QuerySet[Post] | None = None) -> Post:
...
+        context["form"] = form
+        return context
+
+    def post(self: "PostDetailView", request: HttpRequest, pk: int) -> HttpResponse:
+        post = get_object_or_404(Post, id=pk)
+        form = CommentForm(request.POST)
+        if form.is_valid():
+            new_comment = form.save(commit=False)
+            parent_id = request.POST.get("parent_id")
+            if parent_id:
```
jacoor: You can simply set parent_id always and be done with it, no? One if less.

core/templates/core/comments_tree.html
```diff
@@ -0,0 +1,36 @@
...
+                        {% csrf_token %}
+                        {% if form.non_field_errors %}
+                            <div class="alert alert-danger" role="alert">
+                                {% for err in form.non_field_errors %}
+                                    <p class="form-error">{{ err }}</p>
+                                {% endfor %}
+                            </div>
+                        {% endif %}
+                        {{ form|crispy }}
+                        <input type="hidden" name="parent_id" value="{{ comment.id }}">
```
jacoor: Please add it to the form.  
```
from django import forms
from .models import Post

class PostForm(forms.ModelForm):
    extra_field = forms.CharField(max_length=100, required=False)

    class Meta:
        model = Post
        fields = ['title', 'content']  # Specyfikacja pól modelu

```

core/views.py
```diff
@@ -25,6 +28,38 @@ def get_object(self: "Post", queryset: QuerySet[Post] | None = None) -> Post:
...
+        context["comments"] = comments
+        context["form"] = form
+        return context
+
+    def post(self: "PostDetailView", request: HttpRequest, pk: int) -> HttpResponse:
+        post = get_object_or_404(Post, id=pk)
+        form = CommentForm(request.POST)
+        if form.is_valid():
+            new_comment = form.save(commit=False)
+            parent_id = request.POST.get("parent_id")
```
jacoor: This should be in the form, like so
```
from django import forms
from .models import Post

class PostForm(forms.ModelForm):
    extra_field = forms.CharField(max_length=100, required=False)

    class Meta:
        model = Post
        fields = ['title', 'content']  # Specyfikacja pól modelu
```

jacoor (changes requested):
I am missing tests here, could you add?

core/views.py
jacoor: Also, on success it should redirect to view page, so on refresh there will be no "Are you sure you want to resubmit the form" question?

core/templates/core/comments_tree.html
swietlikm: i dont see how i can not use `<input type="hidden" name="parent_id" value="{{ comment.id }}">` in template. How else i can already transfer to from parent_id?

jacoor:
```
from django.shortcuts import render
from .forms import MyForm

def my_view(request):
    initial_data = {
        'name': 'Jan Kowalski',
        'email': 'example@example.com'
    }
    form = MyForm(initial=initial_data)
    return render(request, 'my_template.html', {'form': form})

```

core/models.py
```diff
@@ -14,13 +14,21 @@
 User = get_user_model()
 
 
+class GenericModelManager(models.Manager):
```
jacoor: Please change to ActiveOnlyManager or similar. 
This manager can be used on any model. And current name doesn't reflect what it does.

jacoor (changes requested):

core/templates/core/comments_tree.html
jacoor: This should work - you need to set it in view.

core/views.py
swietlikm: created method for Post model that generate CommentForm, but changed CommentForm from forms.ModelForm to forms.Form, otherwise there would be import loop

jacoor:
A remark about commit namimg
231b953c678cfc81239d0febd1e50eeb786fe3ff
"[updated form handling to set parent_id by default inside form](https://github.com/pymasterspl/reddit/pull/49/commits/231b953c678cfc81239d0febd1e50eeb786fe3ff)" - nope. It also changes this manager. 
Next time let's try to keep commits with functionality.

jacoor (approved):
Looks good, let's see this running.

PR #50, opened 2024-06-18
several post-list fixes

swietlikm:
- fixed queryset of PostListView to .filter(parent=None)
- fixed misspelling in PostListView class
- fixed post list template for better displaying posts, fixed code for better readability

### DIFF
--- a/core/templates/core/post-list.html
+++ b/core/templates/core/post-list.html
@@ -15,7 +15,7 @@
 
             <div class="card-body text-light border">
                 <h3 class="h3 mx-5">{{ post.title }}</h3>
-                <p class="mx-5">{{ post.content|truncatechars:100 }}</p>
+                <p class="mx-5">{{ post.content|striptags|truncatechars:100 }}</p>
             </div>
 
             {% include 'core/post-footer.html' with post=post %}
--- a/core/templates/core/post-list.html
+++ b/core/templates/core/post-list.html
@@ -8,18 +8,21 @@
 {% block content %}
 <div class="container">
     <div class="mx-auto col-10 col-md-8 col-lg-8 mt-lg-3">
-        <a href="{% url 'admin:core_post_add' %}" class="admin-link mb-3">Add New Post</a>
+        <a href="{% url 'admin:core_post_add' %}" class="admin-link mb-3X">Add New Post</a>
         {% for post in posts %}
-            <div class="card bg-dark shadow-sm mb-3 text-light">
-                {% include 'core/post-header.html' with post=post %}
-                    <a href="{% url 'post-detail' post.id %}" class="text-decoration-none">
-                        <div class="card-body text-light">
-                                <h3 class="h3">{{ post.title }}</h3>
-                                {{ post.content|truncatechars:100 }}
-                        </div>
-                    </a>
-                {% include 'core/post-footer.html' with post=post %}
+        <div class="card bg-dark shadow-sm mb-3 text-light border">
+            {% include 'core/post-header.html' with post=post %}
+
+            <div class="card-body text-light border">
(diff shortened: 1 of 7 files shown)


core/templates/core/post-list.html
```diff
@@ -7,34 +7,31 @@
...
-        <li>
-            <div class="card bg-dark shadow-sm mb-2 text-light"
-            >
+            <div class="card bg-dark shadow-sm mb-3 text-light">
                 {% include 'core/post-header.html' with post=post %}
-                <div class="card-body border"
-                     style="cursor: pointer"
-                     id="myDiv{{post.id}}" data-url="{% url 'post-detail' post.id %}"
-                >
+                <div class="card-body border post-clickable" id="post-{{ post.id }}" data-url="{% url 'post-detail' post.id %}">
```
jacoor: This post clickable thing is not so good...
How will it work w/o JS? Also, open in new tab will not work on it cause it is js. So really not good. 
Suggestion: add a traditional link (a) in the h3 below. 
Also, for post content, you need to strip tags. And I would also wrap it in href. This way, we don't need ".,post-clickable" at all. 
Please change. 
Javascript should be used as sugar, extra effects, not to handle core HTML functionality like linking.

jacoor (changes requested):

core/templates/core/post-list.html
```diff
@@ -12,26 +12,15 @@
...
-                <div class="card-body border post-clickable" id="post-{{ post.id }}" data-url="{% url 'post-detail' post.id %}">
-                    <h3 class="h3">{{ post.title }}</h3>
-                    {{ post.content|truncatechars:100 }}
-                </div>
+                    <a href="{% url 'post-detail' post.id %}" class="text-decoration-none">
+                        <div class="card-body text-light">
+                                <h3 class="h3">{{ post.title }}</h3>
+                                {{ post.content|truncatechars:100 }}
+                        </div>
+                    </a>
```
jacoor: Incorrect HTML. 
A can not wrap h3. It needs to be other way around
h3 > a
Div > a

```suggestion
                    
                        <div class="card-body text-light">
                                <h3 class="h3"><a href="{% url 'post-detail' post.id %}" class="text-decoration-none">{{ post.title }}</a></h3>
                                <a href="{% url 'post-detail' post.id %}" class="text-decoration-none">{{ post.content|truncatechars:100 }}</a>
                        </div>
```
You also forgot to add tags stripping to truncate. What will heppen if there will be, ie <b> not closed in those 100 chars?

jacoor (changes requested):

core/templates/core/post-list.html
```diff
@@ -8,18 +8,21 @@
...
-                                {{ post.content|truncatechars:100 }}
-                        </div>
-                    </a>
-                {% include 'core/post-footer.html' with post=post %}
+        <div class="card bg-dark shadow-sm mb-3 text-light border">
+            {% include 'core/post-header.html' with post=post %}
+
+            <div class="card-body text-light border">
+                <h3 class="h3 mx-5">{{ post.title }}</h3>
+                <p class="mx-5">{{ post.content|truncatechars:100 }}</p>
```
jacoor: Still, truncate chars require strip tags or other modifier, to remove html tags before truncate. If the test has any styling (might, in future) it will break.

jacoor (changes requested):
Still one remark.

core/templates/core/post-list.html
jacoor: prob this: https://www.w3schools.com/django/ref_filters_striptags.php

jacoor (approved):

core/templates/core/post-list.html
tgrycuk: done

PR #53, opened 2024-06-20
Feature/account activation via link

Pawelooo:


Pawelooo:
I don't know if poetry.lock is OK because I only made "poetry lock" in cmd and it made a few changes

users/templates/users/account_activation_email.html
```diff
@@ -0,0 +1,19 @@
...
+<head>
+    <meta charset="UTF-8">
+    <title>Register</title>
+</head>
+<body>
+    <h3>Hi, </h3>
+
+    <p>Thank you for registering with Redit Clone. Please click the link below to activate your account:</p>
+
+    <p>{{ domain }}{% url 'activate-account' uidb64=uid token=token %}</p>
```
jacoor: Please double check if this has protocol - http or https.
I think I'd change domain to full server url like https://testserver.com/ 
Mind, http and https has to be dynamically generated.

users/tests/test_activate_account.py
```diff
@@ -0,0 +1,58 @@
...
+    assert not user.is_active
+
+
+@pytest.mark.django_db()
+def test_activate_user_view(client: Client) -> None:
+    password = fake.password()
+    user = User.objects.create_user(nickname="testnickname",
+                                    email="testuser@example.com",
+                                    password=password)
+    user.is_active = False
```
jacoor: Please remove one, it's twice.

users/tests/test_activate_account.py
```diff
@@ -0,0 +1,58 @@
...
+    user.save()
+
+    uid = urlsafe_base64_encode(force_bytes(user.pk))
+    token = account_activation_token.make_token(user)
+
+    activation_url = reverse("activate-account",
+                             kwargs={"uidb64": uid, "token": token})
+    response = client.get(activation_url)
+
+    user.refresh_from_db()
```
jacoor: Refresh from db should be moved lower, before assert user.is_active. 
It's hard to correlate this with checking user.

users/tests/test_activate_account.py
```diff
@@ -0,0 +1,58 @@
...
+    assert response.status_code == 302
+    assert response.url == reverse("login")
+    assert user.is_active
+    messages = list(get_messages(response.wsgi_request))
```
jacoor, line 53: Should not there be only one message? So this should be easy check on message[0]?
and  assert len(messages) ==  1 no?
This way we have full control.

users/templates/users/account_activation_email.html
jacoor: Also, this is not a link - should be <a href...>

users/views.py
```diff
@@ -29,9 +40,29 @@ class UserRegistrationView(FormView):
...
     def form_valid(self: "UserRegistrationView", form: UserRegistrationForm) -> HttpResponse:
-        form.save()
-        email = form.cleaned_data.get("email")
-        messages.success(self.request, f"Account created for {email}!")
+        user = form.save(commit=False)
+        user.is_active = False
+        user.save()
+        uid = urlsafe_base64_encode(force_bytes(user.pk))
+        token = account_activation_token.make_token(user)
+        activation_link = f"http://{self.request.get_host()}/{uid}/{token}/"
```
jacoor: Hmm...
1. What is this activation_link for? I see it's used in send_email, but not in the email template. If you generated it then it should be used in both places. 
2. Also, I am missing here call to "reverse" or "reverse_lazy" to generate this url. 
3. The protocol should be https or http depending if we are on secure server...  An old topic: https://stackoverflow.com/a/21632907

users/views.py
```diff
@@ -43,6 +74,26 @@ def dispatch(self: "LogoutView", request: HttpRequest, *args: tuple, **kwargs: d
         return super().dispatch(request, *args, **kwargs)
 
 
+class ActivateUser(View):
+
+    def get(self: "ActivateUser", request: HttpRequest, uidb64: str, token: str) -> \
+            HttpResponse:
+        try:
+            uid = force_str(urlsafe_base64_decode(uidb64))
+            user = User.objects.get(pk=uid)
```
jacoor: This should check if user is not active.

jacoor (changes requested):
Nice work, couple improvements needed.

users/tests/test_activate_account.py
```diff
@@ -0,0 +1,58 @@
...
+    token = account_activation_token.make_token(user)
+
+    activation_url = reverse("activate-account",
+                             kwargs={"uidb64": uid, "token": token})
+    response = client.get(activation_url)
+
+    assert response.status_code == 302
+    assert response.url == reverse("login")
+    assert not user.is_active
+    user.refresh_from_db()
```
jacoor:
```suggestion
    user.refresh_from_db()
    assert user.is_active
```
This is what I ment. We still need to check if user is active, just keep it in one place.

jacoor (changes requested):
One small change and we are good. Ping on #reddit when ready please.

Pawelooo:
@jacoor You can accept whatever

jacoor (dismissed):

jacoor (approved):
Looks good, but it will hang forever in QA because we need the instructions in other PR

Pawelooo:
I'm already adding documentation to the email

PR #55, opened 2024-06-23
New Community creation

memorisanka:
Hi Guys,
I've added a new community creation form and views.
Now user can see all communities, and add a new community (when logged in).


core/models.py
```diff
@@ -37,6 +36,8 @@ class Meta:
 
 class Community(GenericModel):
     name = models.CharField(max_length=255)
+    slug = models.SlugField(max_length=255, unique=True, blank=True)
+    author = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name='authored_communities')
```
jacoor, line 40: Duplicate field. Generic model has author.

jacoor: Good idea with adding related name

core/models.py
```diff
@@ -159,7 +170,8 @@ def is_saved(self: "Post", user: User) -> bool:
     def get_comments(self: "Post") -> models.QuerySet:
         return self.children.all()
 
-    def get_comment_form(self: "Post") -> CommentForm:
+    def get_comment_form(self: "Post"):
```
jacoor: Why this change? I mean, removing type?

core/templates/core/community-create.html
```diff
@@ -0,0 +1,52 @@
...
+{% endblock %}
+
+{% block content %}
+<style>
+.hidden {
+    display: none;
+}
+
+</style>
+<script>
```
jacoor: I don't see it being used. If it please point me to the place, if not, please remove.

core/tests/test_forms.py
```diff
@@ -0,0 +1,28 @@
...
+    form = CommunityForm(data=form_data)
+
+    assert form.is_valid()
+
+    community = form.save(commit=True)
+
+    expected_slug = slugify(form_data["name"])
+    assert community.slug == expected_slug
+
+    community.save()
```
jacoor: This save seems unnecessary here, no? Community is created in line 17

jacoor (changes requested):

core/models.py
memorisanka: I had a problem with circular imports (models from forms, forms from models), so I moved this import only to import in function. When I moved it, the type throws an error.

core/templates/core/community-create.html
memorisanka: You mean all file is not used?

jacoor: No, I mean this js script.

core/models.py
jacoor: I am aware of it. 
What I pointed out was removing type, not the import.

memorisanka: Because when import was inside the function, the typing gave an error. I don't have an idea how to solve it.

jacoor: oh, interesting. to keep types in place you can add `-> any:` there. Should solve issue and code style is consistent.

core/templates/core/community-create.html
memorisanka: OK, removed. Sorry, I'm not good at frontend and js scripts...

jacoor: Easy, thanks.

core/models.py
memorisanka: No, ruff doesn't allow type Any here! 

obraz

jacoor: image
Sorry, small caps. 
This works in other PR.
Also please remove this TYPE_CHECKING check.

memorisanka: Corrected. Thanks :)

jacoor (dismissed):

core/forms.py
```diff
@@ -37,3 +40,15 @@ def __init__(self: "Post", *args: tuple, **kwargs: dict) -> None:
...
         self.fields["title"].required = True
         self.fields["content"].required = True
+
+
+class CommunityForm(forms.ModelForm):
+    class Meta:
+        model = Community
+        fields: typing.ClassVar = ["name"]
+
+    def __init__(self: "CommentForm", *args: list, **kwargs: dict) -> None:
```
yanazPL: `self: "CommentForm"` vs `self: "CommunityForm"`

core/tests/test_models.py
```diff
@@ -228,3 +229,15 @@ def test_get_saved_posts(user: User, post: Post) -> None:
     assert saved_posts.first().post == post
     SavedPost.remove_saved_post(user=user, post=post)
     assert not saved_posts.exists()
+
+
+@pytest.mark.django_db()
+def test_create_community() -> None:
+    user = User.objects.create_user(
+        email="testuser@example.com", nickname="Test Nickname", password=os.getenv("TEST_PASSWORD")
```
yanazPL: you could use `user` fixture conftest.py - password would be generated on the fly

core/forms.py
memorisanka: Oo, thanks. Corrected!

core/models.py
```diff
@@ -43,6 +46,16 @@ class Meta:
     def __str__(self: "Community") -> str:
         return str(self.name)
 
+    def save(self: "Community", *args: list, **kwargs: dict) -> None:
+        if not self.slug:
```
yanazPL, line 50: this method is simple and elegant

env-template
```diff
@@ -18,4 +18,7 @@ EMAIL_HOST_USER = "test@gmail.com"
 EMAIL_HOST_PASSWORD = "password"
 EMAIL_HOST = "smtp.gmail.com"
 EMAIL_HOST_PORT = 587
-EMAIL_USE_TLS = True
\ No newline at end of file
+EMAIL_USE_TLS = True
+
+# Tests variables
+TEST_PASSWORD = "password"
```
yanazPL: with `user` fixture it is not needed

core/tests/test_models.py
memorisanka: Thanks, I have changed this in the code. Good to know!

env-template
memorisanka: Removed.

yanazPL:
also wondering if we should have more view tests

core/views.py
```diff
@@ -110,3 +111,38 @@ def post(self: "PostSaveView", request: HttpRequest, pk: int, action_type: str)
...
+        response = super().form_valid(form)
+        CommunityMember.objects.create(
+            community=self.object,
+            user=self.request.user,
+            role=CommunityMember.ADMIN,
+        )
+        return response
+
+    def get_success_url(self: "CommunityCreateView") -> str:
+        return reverse_lazy("community-detail", kwargs={"slug": self.object.slug})
```
yanazPL, line 139: another really minor thing, but we can always improve

https://stackoverflow.com/questions/48669514/difference-between-reverse-and-reverse-lazy-in-django

jacoor: What is your suggestion? I mean, I don't see any harm in using reverse_lazy. Even more, to keep it really simple, I'd suggest using reverse_lazy everywhere unless code error. This way we don't have to think: is this a function, attribute of a class, CBV of FBV. 
Simple is best.

yanazPL: Just thought there are reasons why we have two versions of these functions and wanted to use them as intended

yanazPL: But I get it. One variant less confusion

jacoor (approved):

PR #58, opened 2024-06-24
Fix/8694wumg2 wrong link

memorisanka:


core/migrations/0022_auto_20240624_2118.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 5.0.6 on 2024-06-24 19:18
+from __future__ import unicode_literals
+
+from django.db import migrations, transaction
+from django.utils.text import slugify
+
+from core.models import Community
```
jacoor: This wont work. 
As far as I know, you don't access models this way in migrations
`Post = apps.get_model('blog', 'Post')`
https://docs.djangoproject.com/en/5.0/topics/migrations/#data-migrations

core/migrations/0022_auto_20240624_2118.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def unique_slugify(community, slug):
```
jacoor: hmm instead of copying, you can create method create_slug() on Community and call it from the model in the loop below.

jacoor (changes requested):

core/migrations/0022_auto_20240624_2118.py
memorisanka: Done.

core/migrations/0022_auto_20240624_2118.py
memorisanka: Corrected, but as I said, almost all tests failed. Am I missing something?

core/migrations/0022_auto_20240624_2118.py
memorisanka: But...  "when using apps.get_model, you need to avoid calling methods that rely on the current model class directly, such as generate_unique_slug, because this method might not exist or might have a different implementation at the time of the migration." Maybe it should stay as it is?

jacoor (approved):

PR #62, opened 2024-06-27
DO NOT MERGE: Feature/add post delete view

tgrycuk:
### Overview
This pull request adds a new feature to handle the deletion of posts in the application. It includes the necessary view, URL pattern, and template for confirming and executing post deletions.

### Changes
- Added `PostDeleteView` to handle the deletion of posts.
- Updated `urls.py` to include the new URL pattern for post deletion.
- Created a template `post_confirm_delete.html` for the delete confirmation page.

### Reason for Changes
This feature allows users to delete their posts, providing them with more control over their content and enhancing the user experience.

### Testing
- Implemented unit tests to verify that posts are correctly deleted and the user is redirected appropriately.
- Verified that deleted posts are no longer accessible via direct URL or visible in the post list.


core/views.py
```diff
@@ -84,6 +84,16 @@ def get_success_url(self: "PostCreateView") -> str:
         return reverse("post-detail", kwargs={"pk": self.object.pk})
 
 
+class PostDeleteView(DeleteView):
```
swietlikm: This view should be available only for logged users

core/views.py
```diff
@@ -84,6 +84,16 @@ def get_success_url(self: "PostCreateView") -> str:
...
+    model = Post
+    template_name = "core/post-confirm-delete.html"
+    success_url = reverse_lazy("post-list")
```
swietlikm, line 90: Why user is forced to go to post-list? i think it would be better to redirect to the page that was before the button "delete" was pressed

core/views.py
```diff
@@ -84,6 +84,16 @@ def get_success_url(self: "PostCreateView") -> str:
...
+
+    def get_queryset(self: "PostDeleteView") -> models.QuerySet:
+        queryset = super().get_queryset()
+        return queryset.filter(author=self.request.user)
```
swietlikm, line 95: DeleteView is using single objects, this method is not doing anything

swietlikm (changes requested):
right now if i go to post/<int>/delete i can delete any post, also not my posts, i even do not need to be logged to delete someone elses post.
Which task is this PR reffering to?

core/tests/test_views.py
```diff
@@ -69,3 +84,35 @@ def test_add_post_unauthorized(client: Client, community: Community) -> None:
...
+
+    response = client.get(post_detail_url)
+    assert response.status_code == 404
+
+    response = client.get(post_list_url)
+    assert response.status_code == 200
+    assert post.title not in str(response.content)
+    assert "Test Post Titlex" not in str(response.content)
+
+    assert not Post.objects.filter(id=post.id).exists()
```
swietlikm, line 118: i would add another tests to check if other user can delete others posts

jacoor:
@swietlikm there is no task for this PR so it's hanging for tomorrows meeting. 
we need to review and assess implications of this change.

jacoor (changes requested):
This story https://app.clickup.com/t/8694xzk3p is still in discovery. Merge only after the story is ready for work and work done.

core/views.py
tgrycuk: Agreed, I'll ensure this view is accessible only to logged-in users.

core/views.py
tgrycuk: This method is indeed important as it customizes the queryset to ensure that users can only delete posts they authored. This security measure prevents unauthorized deletions, maintaining the integrity of user data.

core/tests/test_views.py
tgrycuk: Great point! However, let's handle this in a separate task to maintain the consistency of our community's policies. Could you please create a follow-up ticket for it?

tgrycuk:
> if i go to post//delete i can delete any post, also not my posts

This statement is not accurate. Please ensure you understand the code in detail before completing the review, particularly the get_queryset method, which filters posts by author to prevent unauthorized deletions. I’m happy to discuss this further if needed.

core/views.py
tgrycuk: done.

PR #64, opened 2024-07-02
Add Avatar Functionality to User Model

tgrycuk:
This pull request adds avatar functionality to the User model. The avatar feature has been implemented and tested to ensure it works correctly, including:

Uploading and resizing avatar images to 32x32 pixels.
Retrieving the correct URL for the user's avatar.
Providing a default avatar URL when no avatar is set.
These changes enhance the user profile experience by allowing users to have personalized avatars.


users/models.py
```diff
@@ -83,14 +104,10 @@ def last_activity_ago(self: "User") -> str:
...
-            result = f"{delta.days} days ago"
-
-        return result
+                return "just now"
+            if delta.seconds < 3600:  # noqa: PLR2004
+                return f"{delta.seconds // 60} minutes ago"
+            return f"{delta.seconds // 3600} hours ago"
+        if delta.days == 1:
+            return "1 day ago"
+        return f"{delta.days} days ago"
```
jacoor, line 113: What happened here? Please revert this change, it was out of scope of your tasks.

users/forms.py
```diff
@@ -9,3 +10,24 @@ class UserRegistrationForm(UserCreationForm):
     class Meta:
         model = User
         fields: ClassVar[list[str]] = ["nickname", "email", "password1", "password2"]
+
+
+class UserProfileForm(forms.ModelForm):
```
jacoor, line 24: Would be good to restrict avatar file size  file size, let's say 2MB.

jacoor (changes requested):

users/models.py
tgrycuk: There was bad solution. Difficult to understand. The new code is significantly easier to read and maintain.

jacoor:
Still, it should be separate commit, at least. Right now it's squeezed into
other work and hard to understand what happened.
One commit, one functionality.
If possible, please revert and push as separate commit.
Dont get me wrong, I like the idea - you are leaving the place better than
found (scout's rule). Still, let's follow the rule of one functionality per
commit.

users/models.py
jacoor: Still, it should be separate commit, at least. Right now it's squeezed into
other work and hard to understand what happened.
One commit, one functionality.
If possible, please revert and push as separate commit.
Dont get me wrong, I like the idea - you are leaving the place better than
found (scout's rule). Still, let's follow the rule of one functionality per
commit.

tgrycuk: I appreciate your feedback and understand the importance of keeping commits focused on a single functionality. However, I believe that good changes should be recognized and accepted, even if they extend beyond the original task scope. Improving the codebase is always beneficial, and these enhancements should be approved.

That said, I agree that the changes should ideally be in separate commits for clarity. I will make sure to follow this practice in the future. For now, let's acknowledge the value of these improvements and approve them as part of this PR.

jacoor: Thank you, agreed.

users/forms.py
jacoor: Still, would be good to limit this uploaded file size. I am pretty sure there is a settings in setting.py we could use. 
I'd like to avoid out of memory issues because of sending huge files.

tgrycuk: done

jacoor (approved):

PR #65, opened 2024-07-05
User related models

yanazPL:
I want to have UserSettings and Profile created by signal when User is created. I've moved nickname from User to Profile and created User.nickname property for backward compability (so every place with User.nickname used works well).

```python
    @property
    def nickname(self):
        return self.profile.nickname if self.profile else "not-yet-named"

    @nickname.setter
    def nickname(self, value):
        if self.profile:
            self.profile.nickname = value
            self.profile.save()
```

I'd like to Profile be created either with username inferred from email or nickname passed to User.

For example here I want to nickname of Profile.nickname be like nickname passed to User:
```python
    user = User.objects.create(email="user@bbb.com", nickname="user1")
    assert user.profile.nickname == "user1"
```

And here to be inffered:

```python
@pytest.mark.django_db()
def test_user_signals():
    user_1 = User.objects.create(email="user@bbb.com")
    user_2 = User.objects.create(email="user@aaa.com")
    # nicknames infered from email
    assert user_1.profile
    assert user_2.profile
    assert user_1.user_settings
    assert user_2.user_settings
    assert user_1.nickname == "user"
    assert user_2.nickname == "user_1"
```

I don't know how to access nickname parameter from inside of User post_save signal. Or maybe I should use some other signal or signals combination?

Pleas look at `signals.py`

Related ClickUp: https://app.clickup.com/t/8694vv3zf

## Summary by Sourcery

This pull request refactors the user-related models by introducing `UserSettings` and `Profile` models, moving the nickname field to the `Profile` model, and adding signals to handle the automatic creation of these models. It also includes updates to the `User` model to support these changes and adds tests to ensure the new functionality works as expected.

- **New Features**:
    - Introduced `UserSettings` and `Profile` models to store user-specific settings and profile information.
    - Added a `gender` field to the `User` model with predefined choices.
    - Implemented a `nickname` property in the `User` model to maintain backward compatibility after moving the nickname to the `Profile` model.
    - Created `SocialLink` model to store social links associated with a user.
- **Enhancements**:
    - Moved the `nickname` field from the `User` model to the `Profile` model.
    - Updated the `User` model to include one-to-one relationships with `UserSettings` and `Profile` models.
    - Added signals to automatically create `Profile` and `UserSettings` instances when a `User` is created.
    - Enhanced the `User` model to infer the nickname from the email if not provided.
- **Tests**:
    - Added tests to verify the creation of `Profile` and `UserSettings` instances upon user creation.
    - Added tests to ensure the nickname is correctly set or inferred for the `Profile` model.


users/models.py
```diff
@@ -94,3 +141,19 @@ def last_activity_ago(self: "User") -> str:
...
 
         return result
+    
+    @property
+    def nickname(self):
+        return self.profile.nickname if self.profile else "not-yet-named"
+
+    @nickname.setter
+    def nickname(self, value):
+        if self.profile:
```
swietlikm, line 151: Since User model is the core model, i would by my intuition put OneToOne field not in User but in Profile and UserSettings. Then I'm wondering if it is not better to modify _create_user method and just add on last 2 lines these: Profile.objects.create(user=user, nickname=nickname)
UserSettings.objects.create(user=user, ... )
then you can remove nickname.setter from User model and you don't need anymore signals. Also since nickname is obligatory during registration im not sure if you need to get it anymore from email

swietlikm:
I have this feeling that since User model is the core model, then i would put onetone fields both inside UserSettings and Profile. Then i would modify _create_user on UserManager and add the last 2 lines as Profile.objects.create(user=user, nickname=nickname) and UserSettings.objects.create(...). Also before _create_user will create the user, you need to check if nickname do not exist yet. This should let you avoid using signals. also nickname.setter can be removed. Also since registration is pushing nickname as obligatory, im not sure if you need to get nickname from email. i guess it was used only for migrations fix.

jacoor:
@yanazPL would be good to update the story, probably this  one: https://app.clickup.com/t/8694r3yxa
with details like this inferring from email

users/models.py
```diff
@@ -46,9 +46,31 @@ def create_superuser(self: "UserManager", email: str, nickname: str, password: s
...
+        max_length= 2,
+        choices= [
+            ("en", "English")
+        ]
+    )
+   #  https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
+    location = models.CharField(
+        max_length=2,
+        choices = [
+            ("US", "United States")
```
jacoor: Might be good to fill it from some source.

users/models.py
```diff
@@ -46,9 +46,31 @@ def create_superuser(self: "UserManager", email: str, nickname: str, password: s
...
+        ]
+    )
+    
+    is_beta = models.BooleanField(default=False)
```
jacoor, line 59: What is this is beta?

users/models.py
```diff
@@ -59,15 +81,40 @@ class User(AbstractUser):
...
-    REQUIRED_FIELDS: ClassVar[list[int]] = ["nickname"]
     objects = UserManager()
     username: None = None
     email: str = models.EmailField(unique=True)
+    gender = models.CharField(choices=GENDER_CHOICES, max_length=1)
     last_activity = models.DateTimeField(auto_now_add=True, db_index=True)
-
-    def __str__(self: "User") -> str:
-        return self.nickname
+    REQUIRED_FIELDS = ["gender"]
```
jacoor: on user? Should it not be on profile?

users/models.py
```diff
@@ -94,3 +141,19 @@ def last_activity_ago(self: "User") -> str:
             result = f"{delta.days} days ago"
 
         return result
+    
+    @property
+    def nickname(self):
```
jacoor: According to reddit, nickname is required so this "not yet named" is not needed.

users/signals.py
```diff
@@ -0,0 +1,27 @@
...
+# Do I need other signals?
+
+@receiver(post_save, sender=User)
+def create_profile(sender, instance, created, **kwargs):
+    if created:
+        profile = Profile.objects.create()
+        if instance.profile and instance.nickname:
+            nickname = instance.profile.nickname
+        else:
+            nickname = nickname_from_email(User, instance)
```
jacoor: No. Reddit has this required upon registration, so should we.

users/signals.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+# Do I need other signals?
+
+@receiver(post_save, sender=User)
+def create_profile(sender, instance, created, **kwargs):
+    if created:
+        profile = Profile.objects.create()
+        if instance.profile and instance.nickname:
+            nickname = instance.profile.nickname
```
jacoor: Wow wait. If there already is instance.profile then there is nothing to do, no?
Especially that nickname has to be required upon registration, as on reddit. 
Reminder: it was inferred from email only for migration.

users/signals.py
```diff
@@ -0,0 +1,27 @@
...
+from .helpers import nickname_from_email
+
+
+# Do I need other signals?
+
+@receiver(post_save, sender=User)
+def create_profile(sender, instance, created, **kwargs):
+    if created:
+        profile = Profile.objects.create()
+        if instance.profile and instance.nickname:
```
jacoor: this instance.profile at this moment will always be false. So only`nickname_from_emai` which is wrong. Details below.

users/signals.py
```diff
@@ -0,0 +1,27 @@
...
+            nickname = instance.profile.nickname
+        else:
+            nickname = nickname_from_email(User, instance)
+        profile.nickname = nickname
+        profile.save()
+        instance.profile = profile
+        instance.save()
```
jacoor:
```suggestion
    if created:
        profile = Profile.objects.create()
        instance.profile = profile
        instance.save()
```
That's all that should be here, I believe.

jacoor (changes requested):
@yanazPL I agree with @swietlikm 
- make username obligatory during registration - so bring it back to the form.  Let's say someone has email like jacek.ostanski@gmail.com - this is privacy data leak and in real world, can lead to fines. Big ones. 
-- Also reddit has username required, no magic around that. 
- we could avoid using one to one field if we use the same ID value on all models. Like have ID field, PK (db change, prob not possible at this moment) and the key would need to be some alphanumeric string, randomly generated. Then if all 3 models have the same it, it is easy to retrieve. But, one to one looks more common. Also, consider this as information, not request.
- as for using signals - good practice task

users/signals.py
yanazPL: Your UserManager doesn't overrwrite create method so signal would work if someone used  UserManager.create()

yanazPL:
> I have this feeling that since User model is the core model, then i would put onetone fields both inside UserSettings and Profile. Then i would modify _create_user on UserManager and add the last 2 lines as Profile.objects.create(user=user, nickname=nickname) and UserSettings.objects.create(...). Also before _create_user will create the user, you need to check if nickname do not exist yet. This should let you avoid using signals. also nickname.setter can be removed. Also since registration is pushing nickname as obligatory, im not sure if you need to get nickname from email. i guess it was used only for migrations fix.

UserManager doesn't overrwrite create method so signal would work if someone used UserManager.create()

users/models.py
yanazPL: > That's all that should be here, I believe.

What should be Profile.nickname be on creation?  You're saying that it should always have value. It will be right after saving form. But before I have to create User. How do I avoid default or "" or None in signal (for training I'm taking signal rather than manager approach [for skills])

jacoor: Understood, this approach is good for learning
1. In this property, just leave "profile not created" or sth. It may happen especially for admin users or old users. 
2. I don't think we need to create user profle on user creation. We should create profile only when user is registered via user registration form, no? This way we will also not create "empty" objects for admin users.

yanazPL: > Understood, this approach is good for learning
> 
>     1. In this property, just leave "profile not created" or sth. It may happen especially for admin users or old users.
> 
>     2. I don't think we need to create user profle on user creation. We should create profile only when user is registered via user registration form, no? This way we will also not create "empty" objects for admin users.

but manual user creation (lest's say with User.objects.create()) won't be functional

yanazPL: aa ok I can change/add  UserManager.. Also is this ok that we have both `.create()` and `.create_user()` in UserManager?

jacoor:
Hey @yanazPL, how are we doing here?

jacoor:
@sourcery-ai review

users/tests/test_user_model.py
```diff
@@ -68,3 +68,25 @@ def test_login_user_view_get(client: Client) -> None:
...
+
+
+@pytest.mark.django_db()
+def test_user_signals():
+    user_1 = User.objects.create(email="user@bbb.com")
+    user_2 = User.objects.create(email="user@aaa.com")
+    # nicknames infered from email
+    assert user_1.profile
+    assert user_2.profile
+    assert user_1.user_settings
```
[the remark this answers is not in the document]
yanazPL: User settings are empty. They can't fail

users/tests/test_user_model.py
```diff
@@ -68,3 +68,25 @@ def test_login_user_view_get(client: Client) -> None:
...
     assert response.status_code == HTTP_SUCCESS
     assert b"Log in" in response.content
+
+
+@pytest.mark.django_db()
+def test_user_signals():
+    user_1 = User.objects.create(email="user@bbb.com")
+    user_2 = User.objects.create(email="user@aaa.com")
+    # nicknames infered from email
+    assert user_1.profile
```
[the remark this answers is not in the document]
yanazPL: Created profiles are empty. It can't fail

users/tests/test_user_model.py
```diff
@@ -68,3 +68,25 @@ def test_login_user_view_get(client: Client) -> None:
...
+    assert user_2.profile
+    assert user_1.user_settings
+    assert user_2.user_settings
+    assert user_1.nickname == "user"
```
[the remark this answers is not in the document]
yanazPL: no nickname inferation from email any more

users/tests/test_user_model.py
```diff
@@ -68,3 +68,25 @@ def test_login_user_view_get(client: Client) -> None:
     response = client.get(reverse_lazy("login"))
     assert response.status_code == HTTP_SUCCESS
     assert b"Log in" in response.content
+
+
+@pytest.mark.django_db()
+def test_user_signals():
+    user_1 = User.objects.create(email="user@bbb.com")
```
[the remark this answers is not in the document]
yanazPL: Is there need to do it? `User.email` is a field with `unique = True`

users/signals.py
yanazPL: There is no nickname creation from email any more

users/signals.py
yanazPL: No nickname inferation any more

users/models.py
yanazPL: On profile now

users/models.py
yanazPL: In reddit it is flag deciding if user wants to user beta version  of reddit. I can remove it if needed.

yanazPL:
Two things:

I'm wondering if `null = True` on `User.profile` and `User.user_settings` and `SocialLink.profile` may be problem.
IDK if I want to deal with Data Migrations right now. Also `profile` and `user_settings` creation is on `post_save` now.

Not sure if avatar should be on `User` not `Profile` It has been made that way.

users/models.py
yanazPL: Should I install some package for it?

users/models.py
```diff
@@ -59,15 +81,40 @@ class User(AbstractUser):
...
         validators=[nickname_validator],
     )
+    bio = models.TextField(default="")
+    is_nfsw_visible = models.BooleanField(default=False)
+    is_followable = models.BooleanField(default=True)
+    is_content_visible = models.BooleanField(default=True)
+    is_communities_visible = models.BooleanField(default=True)
+
+    def __str__(self: "Profile") -> str:
+        return f"{self.nickname} settings"
```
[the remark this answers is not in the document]
yanazPL: Why is it better?

users/models.py
```diff
@@ -59,15 +81,40 @@ class User(AbstractUser):
...
     objects = UserManager()
     username: None = None
     email: str = models.EmailField(unique=True)
+    gender = models.CharField(choices=GENDER_CHOICES, max_length=1)
     last_activity = models.DateTimeField(auto_now_add=True, db_index=True)
-
-    def __str__(self: "User") -> str:
-        return self.nickname
+    REQUIRED_FIELDS = ["gender"]
+    user_settings = models.OneToOneField(UserSettings, related_name="user", on_delete=models.SET_NULL, null=True)
```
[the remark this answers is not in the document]
yanazPL: Why is it wrong?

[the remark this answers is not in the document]
yanazPL: Why is `profile` ok, but not `user_settings`? I can't see a conflict.
```python
user_settings = models.OneToOneField(UserSettings, related_name="user", on_delete=models.SET_NULL, null=True
profile = models.OneToOneField(Profile, on_delete=models.SET_NULL, null=True, related_name="user")
```

users/models.py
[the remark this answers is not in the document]
yanazPL: Isn't it tradeoff? Returning just nickname makes it look the same as `User.__str()__` You know which user it belongs to, but not what model is that

users/models.py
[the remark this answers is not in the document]
yanazPL: @jacoor what do you think?

users/models.py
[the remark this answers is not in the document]
yanazPL: @jacoor what do you think?

users/tests/test_user_model.py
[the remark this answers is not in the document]
yanazPL: Added test above

users/models.py
jacoor: Please. Or just ask chat to get you list. Easier.

users/models.py
jacoor: I am thinking: comment out with a note that it is in original reddit. Right now we don't have use for it.

users/models.py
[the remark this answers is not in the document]
jacoor: Lovely conversation! 
I like 
```
return f"Profile: {self.nickname}"
```

users/models.py
```diff
@@ -59,15 +81,40 @@ class User(AbstractUser):
...
+        OTHER: "Other",
+        NON_SPECIFIED: "Non-specified",
+    }
+
     USERNAME_FIELD = "email"
-    REQUIRED_FIELDS: ClassVar[list[int]] = ["nickname"]
     objects = UserManager()
     username: None = None
     email: str = models.EmailField(unique=True)
+    gender = models.CharField(choices=GENDER_CHOICES, max_length=1)
```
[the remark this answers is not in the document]
jacoor: Given all that rainbow crap we should have undefined there...

users/models.py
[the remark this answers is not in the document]
jacoor: Clarity is really important. 
We could remove related_name no? How it would behave then?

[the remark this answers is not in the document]
jacoor: @yanazPL AI is wrong here. 
We do not need related_name on OneToOne. Please remove.
reference: https://akademiait.com.pl/zrozumiec-django-relacje-pomiedzy-modelami/ 

```
» ./manage.py shell
Python 3.10.7 (main, Sep 15 2022, 01:51:29) [Clang 14.0.0 (clang-1400.0.29.102)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> # import modeli
>>> from accounts.models import StudentCard, Student
>>> # pobranie i wyświetlenie pierwszego studenta
>>> student = Student.objects.first()
>>> student
<Student: John Doe>
>>> # pobranie pierwszej legitymacji
>>> student_card = StudentCard.objects.first()
>>> student_card
<StudentCard: John Doe 123456>
>>> # wyświetlenie studenta z poziomu legitymacji
>>> student_card.student
<Student: John Doe>
>>> student_card.student.first_name
'John'
>>> student_card.student.last_name
'Doe'
>>> # wyświetlenie danych legitymacji z poziomu studenta
>>> # czyli wykorzystanie relacji odwrotnej
>>> student.studentcard
<StudentCard: John Doe 123456>
>>> student.studentcard.card_id
'123456'
>>> student.studentcard.expiration_date
datetime.date(2023, 8, 17)
>>>
```

so it would simply be user.profile and profile.user. 
So please remove related name from profile and user_settings.

jacoor:
@yanazPL your call with data migration, but for data consistency I would prob go for it. 
Check: https://akademiait.com.pl/migracje-w-django-od-podstaw-do-eksperta-czesc-2/ - it's easy

users/signals.py
```diff
@@ -0,0 +1,20 @@
+from typing import Any
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import Profile, User, UserSettings
+
+
+@receiver(post_save, sender=User)
+def create_profile(sender: type, instance: User, created: bool, **kwargs: Any) -> None:  # noqa: ARG001, ANN401, FBT001
```
jacoor: **kwargs: dict
this works so no any and one #noqa less

users/signals.py
```diff
@@ -0,0 +1,20 @@
...
+    if created or not instance.profile:
+        instance.profile = Profile.objects.create()
+        instance.save()
+
+
+@receiver(post_save, sender=User)
+def create_user_settings(sender: type, instance: User, created: bool, **kwargs: Any) -> None:  # noqa: ARG001, ANN401, FBT001
```
jacoor: as above

jacoor (dismissed):
Added couple comments going through the whole discussion, you should get emails. 
ONe small change, not critical so approved.

jacoor:
@yanazPL before you merge it please check if all works
- run project locally
- check if you can register
- add post 
- and so on

yanazPL:
> data migration



> @yanazPL your call with data migration, but for data consistency I would prob go for it. Check: https://akademiait.com.pl/migracje-w-django-od-podstaw-do-eksperta-czesc-2/ - it's easy

I sort of know how to do data migration. I have rather problem with logic then execution. I don't know whom to assign all Profile and UserSettings objects. Should I assign all of them to `User.objects.first()`? Should I create some dummy User which takes all the orphaned Profiles and UserSettings instances (for example if there are no users)? Should I delete orphaned objects?

Also when using RunPython migrations become irreversible unless they are provided with function reversing changes.
Source - https://docs.djangoproject.com/en/5.0/ref/migration-operations/#runpython

users/tests/test_user_model.py
```diff
@@ -135,3 +135,35 @@ def test_user_signals() -> None:
...
+    ):
+        user.user_settings.location = code
+        assert user.user_settings.get_location_display() == location
+
+
+@pytest.mark.django_db()
+def test_non_existing_locaton(user: "User") -> None:
+    user.user_settings.location = "XX"
+    with pytest.raises(ValidationError):
+        user.user_settings.clean_fields()
```
jacoor: Prove that the error is about the location field, please.

users/tests/test_user_model.py
```diff
@@ -135,3 +135,35 @@ def test_user_signals() -> None:
...
+
+
+@pytest.mark.django_db()
+def test_language_location_there_are_some_defaults(user: "User") -> None:
+    loc = user.user_settings.location
```
jacoor: full name please no abbr.

jacoor (changes requested):

jacoor (approved):

PR #86, opened 2024-09-23
Fix: trailing slashes and order for api_vews

yanazPL:
- I added trailing `/` to `api_urls`.
- Changed order of list `api_views` to `-updated_at` - to deal with pytest unexpected order warnings.
- Tried to change it to `-score`, but score is not a field - it's a property.  It cannot be done easily.

## Summary by Sourcery

Fix URL handling by adding trailing slashes to API endpoints and adjust queryset ordering to '-updated_at' to resolve pytest order warnings.

Bug Fixes:
- Add trailing slashes to API endpoint URLs to ensure consistent URL handling.

Enhancements:
- Change the ordering of the PostAPIListView and CommunityPostsListAPIView querysets to order by '-updated_at' to address pytest warnings about unexpected order.

### DIFF
--- a/core/api_views.py
+++ b/core/api_views.py
@@ -8,7 +8,7 @@ from core.serializers import CommunitySerializer, MinimalCommunitySerializer, Po
 
 
 class CommunitiesAPIList(ListAPIView):
-    queryset = Community.objects.exclude(privacy=Community.PRIVATE)
+    queryset = Community.objects.exclude(privacy=Community.PRIVATE).order_by("name")
     serializer_class = MinimalCommunitySerializer
 
 
@@ -24,7 +24,7 @@ class CommunityDetailAPIView(RetrieveAPIView):
 
 
 class PostAPIListView(ListAPIView):
-    queryset = Post.objects.exclude(community__privacy=Community.PRIVATE)
+    queryset = Post.objects.exclude(community__privacy=Community.PRIVATE).order_by("up_votes")
     serializer_class = PostSerializer
 
 
@@ -38,4 +38,4 @@ class CommunityPostsListAPIView(ListAPIView):
         if community.privacy == Community.PRIVATE:
             msg = "Private community is not accessible."
             raise PermissionDenied(msg)
-        return Post.objects.filter(community=community)
+        return Post.objects.filter(community=community).order_by("up_votes")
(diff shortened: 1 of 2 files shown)


core/api_views.py
```diff
@@ -8,7 +8,7 @@
 
 
 class CommunitiesAPIList(ListAPIView):
-    queryset = Community.objects.exclude(privacy=Community.PRIVATE)
+    queryset = Community.objects.exclude(privacy=Community.PRIVATE).order_by("name")
```
jacoor, line 11: Why? I would order by -id to see communities with newest post first.

core/api_views.py
```diff
@@ -24,7 +24,7 @@ def get_serializer_class(self: "CommunityDetailAPIView") -> CommunitySerializer:
 
 
 class PostAPIListView(ListAPIView):
-    queryset = Post.objects.exclude(community__privacy=Community.PRIVATE)
+    queryset = Post.objects.exclude(community__privacy=Community.PRIVATE).order_by("-updated_at")
```
jacoor: I am thinking this should be not -updated_at but by id no? Beceause if we change order -updated here is what happens
- I write post A
- You write post B
Order on page is 
A
B

Now I edit post A

Order will be 
B
A

Am I correct? What are your thoughts?

core/api_views.py
yanazPL: When you look for community usually you want to find specific community. Usually most of them are active.
Also id here would be id of community not posts. That means newer communities are before older communities. This isn't very helpful information when looking for something,

core/api_views.py
yanazPL: I think you are right. Is order of django ids something we can rely on?
There is `created_at` field - that would work the same way, but be more explicit. What do you think?

jacoor: Consider created_at just for displaying purposes - to show date when post are created. 
Id is auto increment - new post, new id, so using ordering by id is something we can rely on. If we use created_at for ordering, we have the same effect, only more DB load. So let's stick to ids.

core/api_views.py
jacoor: ok

jacoor (approved):

PR #98, opened 2024-11-22
Feature/Banner-image

konrad-paszek:
Add banner image feature https://app.clickup.com/t/86965u2ke

## Summary by Sourcery

Add a banner image feature to user profiles, allowing users to upload and display a banner. Update the profile model, forms, and templates to support this feature, and refactor image processing to handle both avatars and banners. Implement tests to validate banner image functionality.

New Features:
- Introduce a banner image feature for user profiles, allowing users to upload and display a banner image on their profile page.

Enhancements:
- Refactor image processing to handle both avatar and banner images using a unified method for resizing and formatting.

Documentation:
- Update user profile and settings templates to include banner image upload and display functionality.

Tests:
- Add tests for banner image processing, including validation for file size and ensuring correct image dimensions after processing.


users/models.py
```diff
@@ -81,15 +91,18 @@ class Profile(models.Model):
...
+    banner = models.ImageField(upload_to=user_banner_path, null=True, blank=True, default=None)
     user = models.OneToOneField("User", on_delete=models.CASCADE, null=False)
 
     def __str__(self: "Profile") -> str:
         return f"{self.user.nickname}"
 
     def save(self: "Profile", *args: any, **kwargs: dict) -> None:
         if self.avatar:
-            self.avatar = self.process_avatar(self.avatar)
+            self.avatar = self.process_image(self.avatar, (32, 32))
```
jacoor, line 103: hmm we might have a bug here. And it was in original implementation. Simply, it was overlooked. 
The file is processed every time when self.avatar or self.banner is set. So, even if the image is not uploaded the image is processed. 
I think, I would use __init__ method to store self._initial_banner and self._initial_avatar to store initial values of banner and avatar, and then check if the values were changed before calling process image in save.

jacoor (changes requested):

users/models.py
konrad-paszek: Corrected :>

jacoor (approved):

PR #100, opened 2024-12-15
Feature/user-api-registration

konrad-paszek:
Add User Api Registration logic: https://app.clickup.com/t/8697356q7

## Summary by Sourcery

Implement a user registration API endpoint with validation and email confirmation, and add corresponding tests.

New Features:
- Introduce a new user registration API endpoint that allows users to register with email, nickname, and password.

Tests:
- Add tests for the user registration API to validate successful registration, missing data handling, and duplicate user scenarios.


users/api_urls.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework.urls import path
+
+from .api_views import UserApiRegistration
+
+urlpatterns = [
+    path("registration/", UserApiRegistration.as_view(), name="api-user-registration"),
```
jacoor: incorrect path. 
Task says users/register, not registration. 
```
Create `POST /users/register` to register a new user.
```

jacoor:
```suggestion
    path("register/", UserApiRegistration.as_view(), name="api-user-registration"),
```

users/api_views.py
```diff
@@ -0,0 +1,15 @@
+from django.template import Context
+from rest_framework.generics import CreateAPIView
+
+from .models import User
+from .serializers import UserSerializer
+
+
+class UserApiRegistration(CreateAPIView):
```
jacoor:
```suggestion
class UserAPIRegistration(CreateAPIView):
```
Let's keep naming consistent with Django.

users/api_views.py
```diff
@@ -0,0 +1,15 @@
...
+    queryset = User.objects.all()
+    serializer_class = UserSerializer
+
+    def get_serializer_context(self: "UserApiRegistration") -> Context:
```
jacoor:
```suggestion
    def get_serializer_context(self: "UserAPIRegistration") -> Context:
```

jacoor:
This is interesting, I did checkout, pull and update. 
95% of tests fail with 
```
FAILED core/tests/test_api.py::test_api_community_list_one_restricted - ModuleNotFoundError: No module named 'magic'
```
[which was introduced in Bug/validation error ([#83 ](https://github.com/pymasterspl/reddit/pull/99)

I did change with importing python-magic, if it breaks project for you, then we need to use diff library.

users/api_views.py
jacoor: On deeper check this method seems unnecessary. I commented it out and tests pass. Please double-check if it is needed. Chat says it does not.

jacoor (changes requested):

users/api_views.py
konrad-paszek: .

users/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+        if attrs["password"] != attrs["password2"]:
+            msg = {"password": "Passwords are not the same", "password2": "Passwords are not the same"}
+            raise serializers.ValidationError(msg)
+        return attrs
+
+    def create(self: "UserSerializer", validated_data: dict) -> User:
+        user = User.objects.create(nickname=validated_data["nickname"], email=validated_data["email"], is_active=False)
+        user.set_password(validated_data["password"])
+        user.save()
+        uid = urlsafe_base64_encode(force_bytes(user.pk))
```
jacoor, line 37: I would move this into utils.py file for users module. The thing is it can and most likely will be shared with "classic" registration.

users/tests/test_api_registration.py
```diff
@@ -0,0 +1,72 @@
...
+        "email": "",
+        "nickname": "",
+        "password": "",
+        "password2": "",
+    }
+    response = client.post(register_url, data)
+    assert response.status_code == 400
+    assert not user_model.objects.filter(email="").exists()
+    for key in data:
+        assert isinstance(response.data[key][0], ErrorDetail)
```
jacoor, line 51: clever. Nice.

jacoor (approved):

PR #102, opened 2024-12-22
Feature/account delete

Richard9306:
According to this task -> https://app.clickup.com/t/86965u2g5

## Summary by Sourcery

Implement account deletion and reactivation functionality.

New Features:
- Users can now delete their accounts.
- Deactivated accounts can be reactivated via email confirmation.

Tests:
- Added tests for account deletion and reactivation.


core/forms.py
```diff
@@ -34,7 +34,7 @@ class Meta:
         widgets: ClassVar[dict[str, forms.Widget]] = {
             "community": forms.Select(attrs={"class": "form-select"}),
             "title": forms.TextInput(attrs={"class": "form-control"}),
-            "content": forms.Textarea(attrs={"class": "form-control"}),
+            "content": forms.Textarea(attrs={"class": "form-control", "maxlength": 1000}),
```
jacoor: Why this limit?
Was this part of the task? Or is it a bug? Database has no limit, this limits the functionality of the whole site.

reddit/settings.py
```diff
@@ -13,6 +13,7 @@
 from pathlib import Path
 
 from decouple import config
+from django.contrib.messages import constants as messages
```
jacoor: I would simply import constants. Is there a name conflict? If not, it will be clearer to not rename.

templates/base.html
```diff
@@ -75,11 +75,11 @@
             </div>
         </nav>
         {% if messages %}
-            <ul class="alert alert-success">
```
jacoor, line 78: PLease do not change tags -> ul vs div. 
Those messages are a list.

templates/base.html
```diff
@@ -75,11 +75,11 @@
...
+            <div class="messages">
                 {% for message in messages %}
-                    <li {% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
```
jacoor, line 80: I don't understand those changes, can you explain? Why was this changed?

users/admin.py
```diff
@@ -23,7 +35,17 @@ class ProfileAdmin(admin.StackedInline):
...
+    list_display: tuple[str] = (
+        "nickname",
+        "email",
+        "is_staff",
+        "is_online",
+        "last_activity_ago",
+        "is_active",
+        "deactivated_at",
+        "reactivate_until",
+        "reactivate_user_link",
```
jacoor: I would not put this on list_display - the table will be hard to read  > too many columns. Do we really need all those fields? At leat reactivate_user_link will be long string, that has no real meaning on the list, in my opinion.

jacoor:
According to readme first line of the PR description should contain link to the task in clickup. Please update.

users/admin.py
```diff
@@ -48,3 +70,51 @@ def is_online(self: "CustomUserAdmin", obj: User) -> bool | str:
...
+        )
+        self.message_user(
+            request,
+            f"The reactivation link has been sent successfully to {user.email} !",
+            messages.SUCCESS,
+        )
+        return redirect("admin:users_user_changelist")
+
+    def reactivate_user_link(self: "CustomUserAdmin", obj: User) -> SafeString | str:
+        if not obj.is_active and obj.reactivate_until is not None:
```
jacoor: How and where is the date verified? This looks like any user can be reactivated, when `reactivate_until` is not none. 
Is that a possible bug? That time check should be done in the above view.

users/forms.py
```diff
@@ -32,6 +36,21 @@ class Meta:
             "is_communities_visible",
         ]
 
+    def validate_file_size(self: "UserProfileForm", file: any, max_size: int, field_name: str) -> None:
```
jacoor: What happened here? This is not part of this task and this is added?
Please explain. This should not be here. 
@konrad-paszek please also verify if this code is good.

users/models.py
```diff
@@ -9,13 +10,16 @@
 from django.contrib.auth.models import AbstractUser
 from django.contrib.auth.validators import UnicodeUsernameValidator
 from django.core.files.base import ContentFile
+from django.core.files.storage import default_storage
```
jacoor, line 13: Hmm.. I am confused, why files here?

users/models.py
```diff
@@ -164,6 +168,20 @@ def banner_url(self: "Profile") -> str:
                 return settings.DEFAULT_BANNER_URL
         return settings.DEFAULT_BANNER_URL
 
+    def delete_avatar(self: "Profile") -> None:
```
jacoor, line 171: Again, I am confused. This is not part of this task. What happened? Please clarify.
@konrad-paszek is this your code? Is this valid? 
My guess it's merge issues but… this is not good.

users/models.py
```diff
@@ -247,6 +267,48 @@ def last_activity_ago(self: "User") -> str:
...
+                self.is_active = False
+                self.nickname = f"deleted_user_{self.pk}"
+                self.email = f"deleted_user_{self.pk}@example.com"
+                self.password = ""
+                self.first_name = ""
+                self.last_name = ""
+                self.is_staff = False
+                self.is_superuser = False
+                self.can_create_post = False
+                self.reactivate_until = None
```
jacoor: I think I would leave this date, and add another one - anonymized_at.. or something, that would keep track of when the account was anonymized. It allows to track how big is the window between when account was ready for anonymization, and when it was really anonymized.
You can also add a flag: is_anonymized to the account, and use it in the condition above. 
it would be like
```python
if not self.is_anonymized:
   second if, like the one you have above
```

users/models.py
```diff
@@ -247,6 +267,48 @@ def last_activity_ago(self: "User") -> str:
...
+
+    def anonymize_related_models(self: "User") -> None:
+        with suppress(UserSettings.DoesNotExist):
+            self.usersettings.delete()
+        try:
+            profile = self.profile
+        except Profile.DoesNotExist:
+            pass
+        else:
+            profile.bio = ""
```
jacoor: Please create a function "anonymize" on profile and call it from here. It will be much more readable.

users/models.py
```diff
@@ -227,6 +245,50 @@ def last_activity_ago(self: "User") -> str:
...
+            profile.is_nsfw = False
+            profile.is_followable = False
+            profile.is_content_visible = False
+            profile.is_communities_visible = False
+            profile.gender = ""
+            profile.user = self
+            profile.delete_avatar()
+            profile.delete_banner()
+            profile.save()
```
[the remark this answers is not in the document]
jacoor: Yea, I wrote same thing.

users/templates/users/account_settings.html
```diff
@@ -3,7 +3,7 @@
         {% include 'users/settings_navbar.html' %}
         <div class="row justify-content-center">
             <div class="col-12 col-md-8">
-                <div class="d-flex justify-content-between"></div>
+<!--                <div class="d-flex justify-content-between"></div> is this div necessary?-->
```
jacoor: Why this change? This div was commented out here but not commented out later in the code - the closing tags. So it probably introduced broken HTML - unclosed div tag.

users/tests/test_user_commands.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from django.contrib.auth import get_user_model
+from django.core.management import call_command
+from django.utils import timezone
+
+User = get_user_model()
+
+
+@pytest.mark.django_db()
+def test_anonymize_inactive_users_command(generated_password: str) -> None:
```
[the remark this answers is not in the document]
jacoor: I think AI is giving us free tests.

users/tests/test_user_form.py
```diff
@@ -0,0 +1,27 @@
...
+@pytest.mark.django_db()
+def test_confirm_password_form_invalid_password(user: User) -> None:
+    wrong_password = f"{user.plain_password}wrong"
+    form = ConfirmDeleteAccountForm(user, data={"password": wrong_password})
+    assert not form.is_valid()
+    assert form.errors["password"] == ["Incorrect password."]
+
+
+@pytest.mark.django_db()
+def test_confirm_password_form_empty_password(user: User) -> None:
```
jacoor, line 24: Nice tests.

users/tests/test_user_model.py
```diff
@@ -262,3 +262,50 @@ def test_cascade_profile_sociallink_deletion(user: "User") -> None:
...
+    user = User.objects.create(
+        nickname="test_user",
+        email="test@example.com",
+        password=generated_password,
+        is_active=True,
+    )
+    user.deactivate()
+    user.refresh_from_db()
+    assert not user.is_active
+    assert user.deactivated_at is not None
```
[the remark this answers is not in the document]
jacoor: Another free test?

users/tests/test_user_model.py
```diff
@@ -262,3 +262,50 @@ def test_cascade_profile_sociallink_deletion(user: "User") -> None:
...
+def test_anonymize_user_successfully(generated_password: str) -> None:
+    user = User.objects.create(
+        nickname="test_user",
+        email="test@example.com",
+        password=generated_password,
+        is_active=False,
+        deactivated_at=timezone.now(),
+        reactivate_until=timezone.now(),
+    )
+    user.anonymize_account()
```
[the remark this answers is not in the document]
jacoor: looks like bunch of free tests!

users/views.py
```diff
@@ -154,3 +157,29 @@ def get_form_kwargs(self: "AccountSettingsView") -> dict[str, Any]:
...
+        form = ConfirmDeleteAccountForm(user=request.user, data=request.POST)
+        if form.is_valid():
+            request.user.deactivate()
+            logout(request)
+            messages.warning(request, "Your account has been deactivated. ")
+            messages.success(
+                request,
+                f"You have {settings.ACCOUNT_EXPIRATION_TIME_IN_DAYS} days to reactivate it."
+                f"After this time, your account will be permanently deleted."
+                f"To reactivate you account, contact our support."
```
jacoor: Ahh that's why admin has the option to send a link to reactivate. 
Curious case:
At the last day of the reactivate period I send message to support to reactivate. They send me the link. But I click it the day later. So I can not reactivate account because the window has passed, right?
This is tricky edge case. Should the reactivation/activation link be valid for certain amount of time? Let's say 48 hours (and this should be stated in the  email)? Let's discuss that, and update clickup task later on, with the conclusion.

jacoor (changes requested):

users/forms.py
konrad-paszek: Yeah, please remove it

users/models.py
konrad-paszek: Yes, please remove it

Richard9306:
> According to readme first line of the PR description should contain link to the task in clickup. Please update.

>Done

core/forms.py
Richard9306: I attempted to fix the issue of negative letter counting in the form.

jacoor: Well, introducing this limit here is major change in functionality, not documented anywhere and done without consultation. This is a great example of why code review is valuable. 
Don't get me wrong - leaving the place in a better shape then found is very good habit. It's just it is the heart of the system and should not be limited w/o discussion. 
If you want to remove the counting issue, the best solution would be to change the field in a way that does not show this counter at all. That solution would also not change functionality.

reddit/settings.py
Richard9306: Done.

core/forms.py
Richard9306: Ok, I fixed this issue in a better way. I added a check for the existence of the maxlength attribute.

templates/base.html
Richard9306: Fixed.

templates/base.html
Richard9306: I don't remember now. I think it was about removing the dot when displaying the message. I've already rolled back the changes.

users/admin.py
Richard9306: Fixed

users/forms.py
Richard9306: Remove the entire method?

users/models.py
Richard9306: Fixed.

users/models.py
Richard9306: This is my code, I added two methods to the Profil model => delete_avatar and delete_banner to delete user files.

users/models.py
Richard9306: I added anonymized_at in the User model.

users/views.py
Richard9306: I've overridden the check_token method in tokens.py to ensure the token does not exceed the reactivate_until time limit.

users/models.py
Richard9306: I needed this to delete user files.

users/admin.py
jacoor: Please remove `reactivate_user_link ` from list display. This link needs to be visible on user edit page, not on users list.  It can be done as a button pretty easily in user-change-form.html. 
https://github.com/fabiocaccamo/django-admin-interface/discussions/262
And official docs: https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#overriding-admin-templates

users/admin.py
```diff
@@ -48,3 +71,51 @@ def is_online(self: "CustomUserAdmin", obj: User) -> bool | str:
...
+        custom_urls = [
+            path(
+                "<int:user_id>/reactivate/",
+                self.admin_site.admin_view(self.reactivate_user_view),
+                name="reactivate_user",
+            )
+        ]
+        return custom_urls + urls
+
+    def reactivate_user_view(self: "CustomUserAdmin", request: HttpRequest, user_id: int) -> HttpResponseRedirect:
```
jacoor: This method also needs a check for valid time range to reactivate. Right now, there might be no link in admin to reactivate user, but it is easy to create /reactivate/ link by hand in browser, which is a security issue.

users/forms.py
jacoor: @Richard9306 please remove this.

users/models.py
jacoor: @Richard9306 please remove this. Along with default_storage import.

users/models.py
```diff
@@ -247,6 +268,48 @@ def last_activity_ago(self: "User") -> str:
...
+
+    def anonymize_related_models(self: "User") -> None:
+        with suppress(UserSettings.DoesNotExist):
+            self.usersettings.delete()
+        with suppress(SocialLink.DoesNotExist):
+            self.profile.sociallink.all().delete()
+        with suppress(Profile.DoesNotExist):
+            self._extracted_from_anonymize_related_models()
+
+    def _extracted_from_anonymize_related_models(self: "User") -> None:
```
jacoor:
```suggestion
    def _anonymize_profile(self: "User") -> None:
```

users/models.py
```diff
@@ -247,6 +268,48 @@ def last_activity_ago(self: "User") -> str:
...
+        self.anonymized_at = timezone.now()
+        self.save()
+
+    def anonymize_related_models(self: "User") -> None:
+        with suppress(UserSettings.DoesNotExist):
+            self.usersettings.delete()
+        with suppress(SocialLink.DoesNotExist):
+            self.profile.sociallink.all().delete()
+        with suppress(Profile.DoesNotExist):
+            self._extracted_from_anonymize_related_models()
```
jacoor:
```suggestion
            self. _anonymize_profile()
```

users/models.py
```diff
@@ -247,6 +268,48 @@ def last_activity_ago(self: "User") -> str:
...
+        self.is_superuser = False
+        self.can_create_post = False
+        self.anonymize_related_models()
+        self.anonymized_at = timezone.now()
+        self.save()
+
+    def anonymize_related_models(self: "User") -> None:
+        with suppress(UserSettings.DoesNotExist):
+            self.usersettings.delete()
+        with suppress(SocialLink.DoesNotExist):
```
jacoor: PLease move this to _anonymize_profile.

users/templates/users/account_settings.html
```diff
@@ -13,7 +13,7 @@ <h2 class="mt-3 mb-3">Account settings</h2>
                                 <h3 class="mt-4 mb-4 text-secondary fs-6 border-bottom border-secondary">Account
                                     preferences</h3>
 
-                                <!-- Email address (manual, not part of form) -->
+                                <!-- Email address (manual, not part of form) ->
```
jacoor: Please revert this change. Valid html comments end with `-->`
THis change here most likely breaks html and for sure removes showing user email field.

users/tests/test_user_model.py
```diff
@@ -280,32 +274,40 @@ def test_deactivate_user_sets_fields_correctly(generated_password: str) -> None:
...
-def test_anonymize_user_successfully(generated_password: str) -> None:
-    user = User.objects.create(
-        nickname="test_user",
-        email="test@example.com",
-        password=generated_password,
-        is_active=False,
-        deactivated_at=timezone.now(),
-        reactivate_until=timezone.now(),
-    )
+def test_anonymize_active_user_does_not_work(user: User) -> None:
```
jacoor: Please update the name to explain why this user was not anonymized. Something like
test_cant_anonymize_active_user or sth.

jacoor (changes requested):
You missed couple earlier places to fix. And some small new comments. 
PLease fix those places.

users/templates/users/account_settings.html
Richard9306: Done

users/models.py
Richard9306: @jacoor Why? I don't understand. Please explain.

users/admin.py
Richard9306: Done

users/forms.py
Richard9306: Done

users/tests/test_user_model.py
Richard9306: Done

users/models.py
Richard9306: Done

users/admin.py
Richard9306: Fixed

users/models.py
jacoor: I still don't see `anonymized_at` field, which would be really nice to have. It will allow us to see whole history of anonymization/deactivation process.

jacoor (changes requested):

pyproject.toml
```diff
@@ -25,6 +25,7 @@ pytest-xdist = "^3.6.1"
 social-auth-app-django = "5.4.1"
 djangorestframework-simplejwt = "^5.3.1"
 drf-yasg = "^1.21.8"
+rest-framework-simplejwt = "^0.0.2"
```
jacoor: @konrad-paszek is this needed? Seems like not added in previous PR? I see it is used, but not pyproject file.

users/models.py
jacoor: @Richard9306 still missing information.

users/models.py
jacoor: Not moved code. After all, it is `profile.sociallink.all()`

jacoor (changes requested):

users/models.py
Richard9306: I use this method for anonymization.

Richard9306: I use this method for anonymization.

pyproject.toml
konrad-paszek: It's not mine

konrad-paszek: IMO we can delete it

jacoor: @Richard9306 please remove if not used.

users/models.py
Richard9306: Should be there now.

users/templates/users/account_settings.html
Richard9306: Fixed.

users/models.py
Richard9306: Fixed.

pyproject.toml
Richard9306: Done.

users/views.py
jacoor: Lets agree on logic
- the link to reactivate account has to be valid for at least 48 hours
- add logic, that upon sending reactivate email, if expiration time is less than now() + 48 hours extends expiration time to now() + 48 hours

jacoor (approved):

PR #104, opened 2024-12-26
Add login/logout API

konrad-paszek:
According to https://app.clickup.com/t/86977e86m

## Summary by Sourcery

Implement login and logout API endpoints and configure JWT authentication.

New Features:
- Added login and logout API endpoints.

Tests:
- Added tests for the new login and logout API endpoints.

### DIFF
--- a/users/api_views.py
+++ b/users/api_views.py
@@ -1,7 +1,7 @@
 from typing import Any, ClassVar
 
 from django.contrib.auth import authenticate, login, logout
-from django.http import HttpRequest
+from requests import Request
 from rest_framework import status
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.exceptions import ValidationError
@@ -22,12 +22,9 @@ class UserAPIRegistration(CreateAPIView):
 
 
 class UserAPILogin(TokenObtainPairView):
-    def post(self: "UserAPILogin", request: HttpRequest, *args: tuple, **kwargs: dict[str, Any]) -> Response:
+    def post(self: "UserAPILogin", request: Request, *args: tuple, **kwargs: dict[str, Any]) -> Response:
         response = super().post(request, *args, **kwargs)
         user = authenticate(username=self.request.data["email"], password=self.request.data["password"])
-        if not user:
-            error_msg = "Incorrect credentials"
-            raise ValidationError(error_msg)
         login(request, user)
         return response
 
@@ -36,7 +33,7 @@ class UserAPILogout(TokenRefreshView):
     authentication_classes: ClassVar[list[str]] = [SessionAuthentication]
     permission_classes = (IsAuthenticated,)
 
-    def post(self: "UserAPILogout", request: HttpRequest, **kwargs: dict[str, Any]) -> Response:  # noqa: ARG002
+    def post(self: "UserAPILogout", request: Request, **kwargs: dict[str, Any]) -> Response:  # noqa: ARG002
         try:
             refresh_token = request.data["refresh"]
             token = RefreshToken(refresh_token)
--- a/users/api_views.py
+++ b/users/api_views.py
@@ -4,7 +4,6 @@ from django.contrib.auth import authenticate, login, logout
 from requests import Request
 from rest_framework import status
 from rest_framework.authentication import SessionAuthentication
-from rest_framework.exceptions import ValidationError
 from rest_framework.generics import CreateAPIView
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
--- a/users/api_views.py
+++ b/users/api_views.py
@@ -1,4 +1,16 @@
+from typing import Any, ClassVar
+
+from django.contrib.auth import authenticate, login, logout
+from django.http import HttpRequest
+from rest_framework import status
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.exceptions import ValidationError
 from rest_framework.generics import CreateAPIView
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework_simplejwt.exceptions import TokenError
+from rest_framework_simplejwt.tokens import RefreshToken
+from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
 
 from .models import User
 from .serializers import UserSerializer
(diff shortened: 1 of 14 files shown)


users/tests/test_api_registration.py
```diff
@@ -1,23 +1,17 @@
...
-    return reverse("api-user-registration")
-
-
 @pytest.mark.django_db()
 def test_registration_form_valid_data(
     client: Client,
     user_model: type[User],
     generated_password: str,
-    register_url: str,
+    api_register_url: str,
```
jacoor, line 14: Please revert. This is for registering via from, not API, as the name of the test states. 
We need new test for API registration.

users/tests/test_api_registration.py
```diff
@@ -37,14 +31,14 @@ def test_registration_form_valid_data(
 
 
 @pytest.mark.django_db()
-def test_registration_form_missing_data(client: Client, user_model: type[User], register_url: str) -> None:
+def test_registration_form_missing_data(client: Client, user_model: type[User], api_register_url: str) -> None:
```
jacoor, line 34: Same as above. Please revert and create specific tests for API, do not break registration tests.

users/tests/test_api_registration.py
jacoor: Or maybe the naming for the tests is completely wrong? If this is designed for API tests then `test_registration_form_valid_data` is at least misleading. In that case, please review all the tests in this file and fix. 
Also, all the tests in this file should be calling API endpoints, not "register_url" or similar.

jacoor: Ah.. on third look.. yes, test names are misleading,  please fix. It should be something like `test_registration_api_valid_data` - not `test_registration_form_valid_data`. You see the issue?

users/tests/test_api_registration.py
jacoor: Misleading test names, please change.

users/api_views.py
```diff
@@ -7,3 +18,26 @@
...
+class UserAPILogin(TokenObtainPairView):
+    def post(self: "UserAPILogin", request: Request, *args: tuple, **kwargs: dict[str, Any]) -> Response:
+        response = super().post(request, *args, **kwargs)
+        user = authenticate(username=self.request.data["email"], password=self.request.data["password"])
+        login(request, user)
+        return response
+
+
+class UserAPILogout(TokenRefreshView):
+    authentication_classes: ClassVar[list[str]] = [SessionAuthentication]
```
jacoor: Wow wait, please explain. Session authentication when we are logging out users by JWT token?
Session auth is cookie based, in API realm, there are no cookies. Session auth is used for "classic" django views, where we use HTTP sites, not for API based views. Or am I not seeing something?

Here is some documentation: https://github.com/jazzband/djangorestframework-simplejwt/blob/9911c85e13eda3f98396923d94342c6f9e5e8d8b/docs/stateless_user_authentication.rst#L24.
JWT auth is stateless, SessionAuth is not stateless. Those are two different authentication methods.

I am thinking we are missing this setting: https://github.com/jazzband/djangorestframework-simplejwt/blob/9911c85e13eda3f98396923d94342c6f9e5e8d8b/docs/getting_started.rst#project-configuration

users/tests/test_api_authentication.py
```diff
@@ -0,0 +1,155 @@
...
+from django.test import Client
+from django.urls import reverse
+from rest_framework.exceptions import ErrorDetail
+from rest_framework_simplejwt.token_blacklist.models import BlacklistedToken, OutstandingToken
+
+User = get_user_model()
+
+
+@pytest.fixture()
+def login_url() -> str:
```
jacoor: Please change names to contain api in the name, like "api_login_url". It will make it more readable.

users/tests/test_api_authentication.py
```diff
@@ -0,0 +1,155 @@
...
+@pytest.fixture()
+def user_credentials(user: User) -> dict[str, str]:
+    return {
+        "email": user.email,
+        "password": user.plain_password,
+    }
+
+
+@pytest.mark.django_db()
+def test_login_valid_data_response_success(
```
jacoor: test_api_login.... for readability please

users/tests/test_api_authentication.py
```diff
@@ -0,0 +1,155 @@
...
+) -> None:
+    response = client.post(login_url, user_credentials)
+    assert response.status_code == 200
+    assert user.is_authenticated
+    assert "access" in response.data
+    assert "refresh" in response.data
+
+
+@pytest.mark.django_db()
+def test_login_invalid_data_response_failed(
```
jacoor: test_api_login...

users/tests/test_api_authentication.py
```diff
@@ -0,0 +1,155 @@
...
+        "password": "12345",
+    }
+    response = client.post(login_url, data)
+    assert response.status_code == 401
+    assert "access" not in response.data
+    assert "refresh" not in response.data
+
+
+@pytest.mark.django_db()
+def test_login_inactive_account_response_failed(
```
jacoor: api in the name..

users/tests/test_api_authentication.py
jacoor, line 1: I am sorry to say but those tests look invalid to me. 
They are based on django client, which is using session authentication, not JWT auth. I don't see any test using JWT call here - in the headers. 
Look at usage section in the documentation: https://github.com/jazzband/djangorestframework-simplejwt/blob/9911c85e13eda3f98396923d94342c6f9e5e8d8b/docs/getting_started.rst#usage. The example with curl shows sending token in the headers. As far as I see you are not putting the token anywhere in the headers. And the switch to session auth in logout serializer only proves it is not using JWT.

users/utils.py
```diff
@@ -0,0 +1,36 @@
...
+from django.utils.encoding import force_bytes
+from django.utils.http import urlsafe_base64_encode
+
+from reddit import settings
+
+from .models import User
+from .tokens import account_activation_token
+
+
+def user_creation(user: User, request: HttpRequest) -> User:
```
jacoor, line 15: good decision.

jacoor (changes requested):

users/tests/test_api_authentication.py
```diff
@@ -90,61 +90,66 @@ def test_refresh_token_valid_data_response_success(
...
 @pytest.mark.django_db()
-def test_logout_valid_data_response_success(
-    client: Client, user: User, login_url: str, logout_url: str, user_credentials: dict[str, str]
+def test_api_logout_valid_data_response_success(
+    client: Client, user: User, api_login_url: str, api_logout_url: str, user_credentials: dict[str, str]
 ) -> None:
-    response = client.post(login_url, user_credentials)
-    refresh_token = response.data["refresh"]
-    response = client.post(logout_url, {"refresh": refresh_token})
+    login_response = client.post(api_login_url, user_credentials)
```
jacoor: Great! 
Now please create a fixture for user logged in via API.
You can also modify django client to send JWT in headers like so:
`client.credentials(HTTP_AUTHORIZATION='Bearer ' + token)`
This will save us couple lines in every test.

jacoor (changes requested):

jacoor (approved):

users/tests/test_api_authentication.py
konrad-paszek: .

PR #106, opened 2025-01-04
add profile/usersettings API

konrad-paszek:
according to -> https://app.clickup.com/t/8697appf2

## Summary by Sourcery

Implement API endpoints for retrieving and updating user profiles and settings.

New Features:
- Added API endpoints for user profile retrieval, user settings retrieval, and user settings update.

Tests:
- Added tests for the new API endpoints.


users/serializers.py
```diff
@@ -3,11 +3,29 @@
...
 
-from .models import User
+from .models import Profile, User, UserSettings
 from .utils import user_creation
 
 
 class UserSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = User
+        fields = "__all__"
```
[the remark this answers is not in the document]
jacoor: Agreed. Using all is bad idea, please use list of fields. 
Rule: Never use __all__.
Explanation: You have a public endpoint for user data. You add some private data to the model, like PESEL, or medical records, etc. Now, those fields are automatically available publicly.
Please change to a list of fields.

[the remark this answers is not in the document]
jacoor: Also, when you see security, please tripple check the comment and it is probably good idea to fix it. If doubts - Discord.

users/api_views.py
```diff
@@ -14,14 +14,39 @@
...
 from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
 
-from .models import User
-from .serializers import UserSerializer
+from .models import Profile, User, UserSettings
+from .serializers import ProfileSerializer, UserRegistrationSerializer, UserSerializer, UserSettingsSerializer
 from .tokens import account_activation_token
 
 
+class JWTAuthenticatedView:
```
jacoor: Isn't this automatically supported by REST framework?
image
At least authentication_classes is not needed, so only permission_classes should be used. In that case, I would change JWTAuthenticatedView to AuthenticatedView, move to utis.py file  and use in both, views and API views.

users/api_views.py
```diff
@@ -14,14 +14,39 @@
...
+    serializer_class = UserRegistrationSerializer
+
+
+class UserRetrieveAPIView(JWTAuthenticatedView, RetrieveAPIView):
     queryset = User.objects.all()
     serializer_class = UserSerializer
+    lookup_field = "nickname"
+
+
+class ProfileAPIView(JWTAuthenticatedView, RetrieveAPIView, UpdateAPIView):
```
jacoor: Please change name so it is clear it is retrieve logged in user. Like CurrentUserProfileAPIView or sth.

users/api_views.py
```diff
@@ -14,14 +14,39 @@
...
+    serializer_class = ProfileSerializer
+
+    def get_object(self: "UserSettingsAPIView") -> UserSettings:
+        return Profile.objects.select_related("user").get(user__id=self.request.user.id)
+
+
+class UserSettingsAPIView(JWTAuthenticatedView, RetrieveAPIView, UpdateAPIView):
```
jacoor: Same as above.

users/serializers.py
```diff
@@ -1,34 +1,65 @@
...
+class ProfileSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Profile
+        fields = "__all__"
+
+
+class UserSettingsSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = UserSettings
+        fields = "__all__"
```
jacoor: Do not use __all__. Ever. Explained in another place.

users/serializers.py
```diff
@@ -1,34 +1,65 @@
...
+            "email",
+            "avatar",
+            "user_bio",
+            "banner",
+            "post_karma",
+            "comment_karma",
+        ]
+
+
+class ProfileSerializer(serializers.ModelSerializer):
```
jacoor, line 29: Do no use __all__. Ever. 
Explanation: https://github.com/pymasterspl/reddit/pull/106#discussion_r1903133129

users/serializers.py
```diff
@@ -1,34 +1,65 @@
...
         user = User.objects.create(nickname=validated_data["nickname"], email=validated_data["email"], is_active=False)
         user.set_password(validated_data["password"])
         user.save()
         request = self.context.get("request")
         user_creation(user, "api-activate-account", request)
         return user
 
-    def get_message(self: "UserSerializer", obj: User) -> str:
+    def get_message(self: "UserRegistrationSerializer", obj: User) -> str:
         return f"Account created for {obj}! " f"Please confirm your email to activate " f"your account."
```
jacoor: What is this weird formatting? Why not one f string? I know, it's old.. Just curious.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+    return reverse("api-user-profile")
+
+
+@pytest.fixture()
+def api_user_settings_url() -> str:
+    return reverse("api-user-settings")
+
+
+@pytest.mark.django_db()
+def test_api_user_valid_data_response_success(
```
jacoor: A bit of naming issue. 
The tests is checking GET method. But the name is confusing and for me suggests it's updae. 
Please change the name of each of the test to clearly state is it get/patch/post(create) or other. 
THe problem is I need to read carefully test code to see it GET. So when I am seeking for the test to GET, I can't use collapsed view, I need to have all the code listed and read it through, which is a waste of time. 
Thank you.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+@pytest.mark.django_db()
+def test_patch_api_user_profile_valid_data_response_success(
+    authenticated_client: tuple[APIClient, str, str], api_user_profile_url: str, user: User
+) -> None:
+    client, _, _ = authenticated_client
+    data = {"is_followable": not user.profile.is_followable}
+    response = client.patch(api_user_profile_url, data=data)
+    assert response.status_code == 200
+    assert response.data == {
+        "id": user.profile.id,
```
jacoor: Just a remark, about testing. 
As a good practice it is better to use values than variables in assertions in testing. Of course, it varies case by case. But, for exampe adding a string inot "bio" like
```
"bio" : "this is example user bio, lorem ipsum etc"
```
shows more detail, not assumptions. You can clearly see it is a string, not null for example. 
I am not saying anything is to change here. Just saying to always consider using real values, not variables, cause it helps in profile readability. 
It may lower a bit test maintainability, though, so it's always a tradeoff.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+    authenticated_client: tuple[APIClient, str, str], api_user_profile_url: str, user: User
+) -> None:
+    client, _, _ = authenticated_client
+    data = {"is_followable": "TEST"}
+    response = client.patch(api_user_profile_url, data=data)
+    assert response.status_code == 400
+
+
+@pytest.mark.django_db()
+def test_api_user_profile_unauthenticated_user_response_failed(
```
jacoor: missing get in name. It's inconsistency in naming. In the test in like 105 it states clearly patch, this does not contain method name in test name.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+
+@pytest.mark.django_db()
+def test_patch_api_user_settings_valid_data_response_success(
+    authenticated_client: tuple[APIClient, str, str], api_user_settings_url: str, user: User
+) -> None:
+    client, _, _ = authenticated_client
+    data = {"is_beta": not user.usersettings.is_beta}
+    response = client.patch(api_user_settings_url, data=data)
+    assert response.status_code == 200
+    assert response.data["is_beta"]
```
jacoor: Hmm actually I don't see a change here. It has some value, it's not showing that the call changed anything. Please prove that is_beta was changed.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+@pytest.mark.django_db()
+def test_patch_api_user_settings_invalid_data_response_failed(
+    authenticated_client: tuple[APIClient, str, str], api_user_settings_url: str, user: User
+) -> None:
+    client, _, _ = authenticated_client
+    data = {"is_beta": "TEST"}
+    response = client.patch(api_user_settings_url, data=data)
+    assert response.status_code == 400
```
jacoor, line 169: Please prove that the error is only on is_beta. This shows only that there is an error, we don't know what is it.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+
+@pytest.mark.django_db()
+def test_patch_api_user_settings_is_beta_response_success(
+    authenticated_client: tuple[APIClient, str, str], api_user_settings_url: str, user: User
+) -> None:
+    client, _, _ = authenticated_client
+    data = {"is_beta": not user.usersettings.is_beta}
+    response = client.patch(api_user_settings_url, data=data)
+    assert response.status_code == 200
+    assert response.data["is_beta"] == (not user.usersettings.is_beta)
```
jacoor: Duplicate?
test_patch_api_user_settings_valid_data_response_success.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+@pytest.mark.django_db()
+def test_patch_api_user_settings_content_lang_response_success(
```
jacoor: I am thinking it would be better to merge those tests into one, because of running time. 
Each of the tests has a tradeoff: set up db, clean up afterwards. It takes time. 
This can be easily merged into one test which will shorten tests running time.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+    authenticated_client: tuple[APIClient, str, str], api_user_settings_url: str, user: User
+) -> None:
+    client, _, _ = authenticated_client
+    data = {"content_lang": "es"}
+    response = client.patch(api_user_settings_url, data=data)
+    assert response.status_code == 200
+    assert response.data["content_lang"] == "es"
```
jacoor: This does not prove any change happened. 
Hint: what was value before the call? Please update all tests.

users/tests/test_api_user.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+@pytest.mark.django_db()
+def test_patch_api_user_settings_is_over_18_response_success(
+    authenticated_client: tuple[APIClient, str, str], api_user_settings_url: str, user: User
+) -> None:
+    client, _, _ = authenticated_client
+    data = {"is_over_18": not user.usersettings.is_over_18}
```
jacoor: This kind of proves the change but does not prove it.. 
Consider this|
image
Please update other places too.

jacoor (changes requested):

jacoor (approved):

PR #114, opened 2025-01-18
Feature/saved posts on profile

konrad-paszek:
Add saved-posts list to the profile according to this ticket -> https://app.clickup.com/t/8694thy5x

## Summary by Sourcery

New Features:
- Display a list of saved posts on a dedicated tab in the user profile.


core/views.py
```diff
@@ -37,6 +37,16 @@ def get_queryset(self: "PostListView") -> models.QuerySet:
         return Post.objects.filter(parent=None, is_active=True)
 
 
+class SavedPostListView(LoginRequiredMixin, ListView):
+    template_name = "core/post-list-saved.html"
+    context_object_name = "posts"
+
+    def get_queryset(self: "SavedPostListView") -> models.QuerySet:
+        return Post.objects.filter(
+            id__in=SavedPost.objects.filter(user=self.request.user).values("post"), parent=None, is_active=True
```
jacoor: I am not sure about this parent "None" here. It wont show a post from thread.. What if the thread is very long and I want to save only post #99?

users/templates/users/profile.html
```diff
@@ -1,36 +1,32 @@
...
-                    {% for award in user.awards_received.all %}
-                        <img src="{% static 'core/icons/' %}{{ award.choice }}.png" alt="{{ option.label }}" width="30" height="30">
-                    {% endfor %}
-                </p>
-                <a href="{% url 'account_settings' %}" class="btn btn-primary">Edit Profile</a>
-            </div>
-        </div>
-    </div>
-{% endblock %}
+<ul class="nav nav-pills mb-3 justify-content-center custom-nav-pills">
```
jacoor, line 5: Will those links ever be visible? Block "content_profile" is basically filling block content if you look at change in base.html. 
I did not run this, but it looks weird. Possible bug. 
Please make a test: move block "content_profile" out of "content" in this file, put some content in "content_profile" and check how this site renders. 
If I am right, you will only see content from "content_proffle".

users/urls.py
```diff
@@ -19,6 +21,7 @@
     path("profile/edit/profile/", ProfileSettingsView.as_view(), name="profile_settings"),
     path("profile/edit/account/", AccountSettingsView.as_view(), name="account_settings"),
     path("profile/delete-account", AccountDeleteView.as_view(), name="delete_account"),
+    *profile_urlpatterns,
```
jacoor, line 26: Please explain this notation, I see it for the first time. 
I would expect `+` at the end, not `*`.
Link to documentation welcome.

jacoor (commented):
Some clarifications please. Looks good otherwise.

core/views.py
konrad-paszek: I don't understand, this is mechanism to list saved post no saving a post, this is already done in the code

users/urls.py
konrad-paszek: Ruff told me to this way, I wanted use +, but ruff "users\urls.py:19:15: RUF005 Consider iterable unpacking instead of concatenation
Found 2 errors (1 fixed, 1 remaining)."

users/templates/users/profile.html
konrad-paszek: It won't work in this way, only in this way which I have done

konrad-paszek: Links are visible

core/views.py
jacoor: if post has parent then `parent=None,` will make it skip. So it will not be shown, right?

jacoor: I saved last post, it is not visible on list in profile
image
It is, when I remove this None mentioned above.

Please fix query.

templates/base.html
```diff
@@ -83,6 +83,9 @@
         {% endif %}
 
         {% block content %}
+            {% block content_profile %}
```
jacoor: this change is not needed - profile works properly w/o it, please remove

users/templates/users/profile.html
jacoor: right, https://github.com/pymasterspl/reddit/pull/114/files#diff-690b81fad8df2a1f1ce37c846641d9247e3472d244f17039d2910a5ed0c98d5eR86 this change is not needed,. Just checked.

users/urls.py
jacoor: aaah clever unpacking. Nice!

jacoor (changes requested):
Please fix issues.

jacoor (approved):

PR #121, opened 2025-01-27
User Posts List

Revoxer:
Task: https://app.clickup.com/t/8697hf2qj

## Summary by Sourcery

Add the ability for users to list, edit, and delete their own posts.

New Features:
- Introduce views, forms, and templates for managing user posts.
- Add filtering options to the user post list view to display published, draft, and archived posts.
- Implement "edit" and "delete" functionality for user-owned posts.

Tests:
- No tests were added as part of this change.


core/profile_urls.py
```diff
@@ -1,7 +1,10 @@
 from django.urls import path
 
-from core.views import SavedPostListView
+from core.views import SavedPostListView, UserPostDeleteView, UserPostEditView, UserPostListView
 
 profile_urlpatterns = [
     path("profile/saved-post-list/", SavedPostListView.as_view(), name="saved_posts"),
+    path("profile/posts/", UserPostListView.as_view(), name="user_posts"),
+    path("profile/posts/edit/<int:pk>/", UserPostEditView.as_view(), name="edit_post"),
+    path("profile/posts/delete/<int:pk>/", UserPostDeleteView.as_view(), name="delete_post"),
```
konrad-paszek: You created a whole logic for editing and deleting post, nice! But let's move those options to core urls instead of profile and also to complete the full logic let's use those urls also for post list in main page

core/views.py
```diff
@@ -46,6 +47,49 @@
...
+class UserPostDeleteView(LoginRequiredMixin, DeleteView):
+    model = Post
+    template_name = "core/post-delete.html"
+    context_object_name = "post"
+    success_url = reverse_lazy("user_posts")
+
+    def get_queryset(self: "UserPostDeleteView") -> models.QuerySet:
+        return Post.objects.filter(author=self.request.user)
 
 @method_decorator(login_required, name="post")
```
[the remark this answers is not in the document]
konrad-paszek: Please add unit tests for your logic

konrad-paszek (changes requested):

core/forms.py
```diff
@@ -44,6 +44,21 @@ def __init__(self: "Post", *args: tuple, **kwargs: dict) -> None:
         self.fields["content"].required = True
 
 
+class PostUpdateForm(forms.ModelForm):
```
jacoor: This looks like a good place to practice inheritance. 
In my opinion:
- PostUpdateForm should be the parent of PostForm
- meta would probably need to stay as is in both forms (I don't recall option to update part of Meta - good case for quick research - You may check AI)
- __init__ should be inherited and expanded. So in PostForm it would be
```
   def __init__(self: "Post", *args: tuple, **kwargs: dict) -> None:
        super().__init__(*args, **kwargs)
        self.fields["community"].queryset = Community.objects.filter(is_active=True)
```
Your call if you want to try this :-)

core/models.py
```diff
@@ -182,6 +182,9 @@ class Post(GenericModel):
     )
     display_counter = models.IntegerField(default=0)
     is_active = models.BooleanField(default=True)
+    is_draft = models.BooleanField(default=False)
+    is_archive = models.BooleanField(default=False)
+    is_published = models.BooleanField(default=False)
```
jacoor, line 188:
```suggestion
    # FIXME: those fields (is_active, is_draft, is...) need to be put into one field to avoid logic hell later on.
    is_active = models.BooleanField(default=True)
    is_draft = models.BooleanField(default=False)
    is_archive = models.BooleanField(default=False)
    is_published = models.BooleanField(default=False)

```

Please commit suggestion, I just want a comment there because those multiple fields will lead to very complicated logic in future. We need to properly desing this. Right now it could have is_published = true and is_draft = true, which does not make sense.

core/templates/core/user-post-list.html
```diff
@@ -0,0 +1,83 @@
+{% extends 'users/profile.html' %}
+{% load static %}
+
+{% block extra_css %}
```
jacoor: I think I would throw it into users/profile.html.  So we don't need to add it on those subpages. 
Your call.

core/views.py
```diff
@@ -47,6 +48,54 @@ def get_queryset(self: "SavedPostListView") -> models.QuerySet:
...
 
 
+class UserPostListView(LoginRequiredMixin, ListView):
+    template_name = "core/user-post-list.html"
+    context_object_name = "posts"
+
+    def get_queryset(self: "UserPostListView") -> models.QuerySet:
+        user_post = Post.objects.filter(author=self.request.user).order_by("-created_at")
+        status = self.request.GET.get("status")
+        if status == "draft":
```
jacoor: Please use match case: 
https://akademiait.com.pl/jak-zaimplentowac-switch-case-w-pythonie/ section: Python 3.10 i dopasowywanie wzorców strukturalnych.

jacoor: Or you could use a dict, which is bit more complicated but would be much shorter syntax.

jacoor (changes requested):
Couple small changes. This looks good!

core/views.py
```diff
@@ -47,6 +48,55 @@ def get_queryset(self: "SavedPostListView") -> models.QuerySet:
         )
 
 
+class UserPostListView(LoginRequiredMixin, ListView):
+    template_name = "core/user-post-list.html"
+    context_object_name = "posts"
+
+    def get_queryset(self: "UserPostListView") -> models.QuerySet:
+        user_post = Post.objects.filter(author=self.request.user).order_by("-created_at")
```
jacoor: We have a bug here. 
Everything is a Post object in the system.
A post with parent pointing to other post is comment. A post with parent = null is Post - start of a thread. So please add here condition to check parent = null. 
Look here: https://github.com/pymasterspl/reddit/blob/08e1ae4566f1f5c5cfb6aed5c09f0db3a0a41697/core/models.py#L38

jacoor (changes requested):

core/templates/core/post-edit.html
```diff
@@ -0,0 +1,43 @@
...
+            {% endfor %}
+        </ul>
+    </div>
+    {% endif %}
+
+    <form method="post">
+        {% csrf_token %}
+
+        <div class="mb-3">
+            <label for="{{ form.title.id_for_label }}" class="form-label">{{ form.title.label }}</label>
```
jacoor, line 21: For the record: this is just a remark, not a change request. This code is good. 

Have you maybe checked other forms?
The way you do it here is good, however going field over field with copy/paste is bad practice and usually a busywork.
We typically use https://github.com/pymasterspl/reddit/blob/08e1ae4566f1f5c5cfb6aed5c09f0db3a0a41697/core/templates/core/post-detail.html#L53 - form|crispy and it does the job. 

If you really want to go field over field, and interesting approach is to loop over form.fields, to minimize code repetition.

core/templates/core/user-post-list.html
```diff
@@ -0,0 +1,78 @@
...
+            <div class="card-body text-light border">
+                <h3 class="h3 mx-5">{{ post.title }}</h3>
+                <p class="mx-5">{{ post.content|striptags|truncatechars:100 }}</p>
+            </div>
+            
+        </div>
+        <div class="d-flex flex-row-reverse">
+            <a href="{% url 'post-detail' post.id %}" class="button text-decoration-none h3 me-5 mb-5">Read More</a>
+        </div>
+        {% empty %}
```
jacoor, line 71: Nice!

users/tests/test_post.py
```diff
@@ -0,0 +1,52 @@
...
+from django.test import Client
+from django.urls import reverse
+
+from core.models import Post
+
+User = get_user_model()
+
+
+@pytest.mark.django_db()
+def test_edit_post_title(client: Client, user: User, post: Post) -> None:
```
jacoor, line 12: Thank you! Great to have tests!

jacoor (approved):

konrad-paszek (approved):

PR #125, opened 2025-02-04
Upvote content list

Revoxer:
Task: https://app.clickup.com/t/8697hf6p5

## Summary by Sourcery

New Features:
- Allow users to view their upvoted posts and comments.

### DIFF
--- a/core/templates/core/user-upvoted.html
+++ b/core/templates/core/user-upvoted.html
@@ -54,7 +54,6 @@
                             </a>
                             <ul class="dropdown-menu dropdown-menu-dark">
                                 <li>
-                                    {% if request.user.is_authenticated %}
                                     {% if post|is_saved:request.user %}
                                     <form method="POST" class="d-none" id="unsave-form-{{ post.pk }}" action="{% url 'post-save-unsave' pk=post.pk action_type='unsave' %}?next={{ request.path }}#post-{{post.id}}">
                                         {% csrf_token %}
@@ -66,17 +65,12 @@
                                     </form>
                                     <a href="#" class="dropdown-item" onclick="event.preventDefault(); document.getElementById('save-form-{{ post.pk }}').submit();"><i class="bi bi-bookmark"></i> Save</a>
                                     {% endif %}
-                                    {% else %}
-                                    <a href="{% url 'login' %}?next={{ request.path }}" class="dropdown-item"><i class="bi bi-bookmark"></i> Save</a>
-                                    {% endif %}
                                 </li>
                                 <li><a class="dropdown-item" href="#"><i class="bi bi-eye-slash"></i> Hide</a></li>
                                 <li><a class="dropdown-item" href="{%  url 'post-report' post.id %}"><i class="bi bi-flag"></i> Report</a></li>
-                                {% if request.user.is_authenticated %}
                                 {% if post.author == request.user %}
                                 <li><a class="dropdown-item" href="{% url 'edit_post' post.pk %}"><i class="bi bi-pencil-square"></i> Edit</a></li>
                                 <li><a class="dropdown-item" href="{% url 'delete_post' post.pk %}"><i class="bi bi-trash"></i> Delete</a></li>
-                                {% endif %}
                                 {% endif %}
                             </ul>
                         </div>
--- a/core/templates/core/user-upvoted.html
+++ b/core/templates/core/user-upvoted.html
@@ -1,0 +1,104 @@
+{% extends 'users/profile.html' %}
+{% load is_saved %}
+
+{% block content_profile %}
+<div class="container">
+    <div class="mx-auto col-10 col-md-8 col-lg-8 mt-lg-3">
+        <div class="mb-3">
+            <span>Filter as:</span>
+            <form method="get" class="d-inline-block ms-2">
+                <select name="filter" onchange="this.form.submit()" class="form-select bg-dark text-light">
+                    <option value="" {% if not request.GET.filter %}selected{% endif %}>All</option>
+                    <option value="post" {% if request.GET.filter == "post" %}selected{% endif %}>Post</option>
+                    <option value="comment" {% if request.GET.filter == "comment" %}selected{% endif %}>Comment</option>
+                </select>
+            </form>
+        </div>
+
+        {% for post in posts %}
+        <div id="post-{{ post.id }}" class="card bg-dark shadow-sm mb-3 text-light border">
+            
+            <div class="card-header">
+                <div class="row align-items-center">
+                    <div class="col-auto">
+                         <img src="{{ post.author.profile.avatar_url }}"
+                               alt="{{ post.author }} avatar image"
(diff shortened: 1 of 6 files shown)


core/templates/core/user-upvoted.html
```diff
@@ -0,0 +1,104 @@
...
+                        </div>
+                    </div>
+                    <div class="col-auto ms-auto">
+                        <div class="dropdown bg-dark">
+                            <a class="btn btn-dark" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+                                <i class="bi bi-three-dots"></i>
+                            </a>
+                            <ul class="dropdown-menu dropdown-menu-dark">
+                                <li>
+                                    {% if request.user.is_authenticated %}
```
jacoor: It's always will be authenticated here. Not needed.

core/templates/core/user-upvoted.html
```diff
@@ -0,0 +1,104 @@
...
+                                    </form>
+                                    <a href="#" class="dropdown-item" onclick="event.preventDefault(); document.getElementById('save-form-{{ post.pk }}').submit();"><i class="bi bi-bookmark"></i> Save</a>
+                                    {% endif %}
+                                    {% else %}
+                                    <a href="{% url 'login' %}?next={{ request.path }}" class="dropdown-item"><i class="bi bi-bookmark"></i> Save</a>
+                                    {% endif %}
+                                </li>
+                                <li><a class="dropdown-item" href="#"><i class="bi bi-eye-slash"></i> Hide</a></li>
+                                <li><a class="dropdown-item" href="{%  url 'post-report' post.id %}"><i class="bi bi-flag"></i> Report</a></li>
+                                {% if request.user.is_authenticated %}
```
jacoor: User is always authenticated here, not needed.

jacoor (changes requested):
A test would be nice. Looks good. If you want - add couple tests. And please remove this not needed is authenticated from template.

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@pytest.mark.django_db()
+def test_upvoted_post(client: Client, user: User, post: Post, upvote_post: PostVote) -> None:
+    url = reverse("user_upvoted")
+    client.force_login(user)
+    response = client.get(url)
+    post.refresh_from_db()
+    upvote_post.refresh_from_db()
+    assert response.status_code == 200
```
jacoor, line 16: This test is not good. It only tests if response is 200. 
The assertions below are only checking DB data, no from the retrieved page. 
You need to check `response.context` - it is the object with context data for the page. So the posts to check are most likely in response.context.posts.  Given that, the test should be something like
```
assert len(response.context.posts) == 1 # one upvoted post. 
# access the rest of data in similar way
```

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+    assert upvote_post.choice == PostVote.UPVOTE
+    assert post.up_votes == 1
+
+
+@pytest.mark.django_db()
+def test_upvoted_comment(client: Client, user: User, comment: Post, upvote_comment: PostVote) -> None:
+    url = reverse("user_upvoted")
+    client.force_login(user)
+    response = client.get(url)
+    comment.refresh_from_db()
```
jacoor: Same as above. Refresh from DB is not needed.

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+    assert upvote_comment.choice == PostVote.UPVOTE
+    assert comment.up_votes == 1
+
+
+@pytest.mark.django_db()
+def test_filter_post(client: Client, user: User, post: Post, upvote_post: PostVote) -> None:
+    url = reverse("user_upvoted") + "?filter=post"
+    client.force_login(user)
+    response = client.get(url)
+    post.refresh_from_db()
```
jacoor: Not needed.

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+    upvote_post.refresh_from_db()
```
jacoor: Not needed.

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+    assert response.status_code == 200
+    assert post in response.context["posts"]
```
jacoor: yup, that is the way to test context. 
But, prove me that there is only one post in the response, please.

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@pytest.mark.django_db()
+def test_filter_comment(client: Client, user: User, comment: Post, upvote_comment: PostVote) -> None:
+    url = reverse("user_upvoted") + "?filter=comment"
+    client.force_login(user)
+    response = client.get(url)
+    comment.refresh_from_db()
+    upvote_comment.refresh_from_db()
```
jacoor:
```suggestion

```
not needed.

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+    assert response.status_code == 200
+    assert comment in response.context["posts"]
```
jacoor: Prove me only this comment in response.

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@pytest.mark.django_db()
+def test_filter_all_post(client: Client, user: User, post: Post, upvote_post: PostVote) -> None:
+    url = reverse("user_upvoted") + "?filter="
+    client.force_login(user)
+    response = client.get(url)
+    post.refresh_from_db()
+    upvote_post.refresh_from_db()
```
jacoor:
```suggestion

```

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+    assert response.status_code == 200
+    assert post in response.context["posts"]
```
jacoor: Prove me please there are boths, posts and comment. And only 2.

users/tests/test_vote_content.py
```diff
@@ -0,0 +1,78 @@
...
+
+@pytest.mark.django_db()
+def test_filter_all_comment(client: Client, user: User, comment: Post, upvote_comment: PostVote) -> None:
+    url = reverse("user_upvoted") + "?filter="
+    client.force_login(user)
+    response = client.get(url)
+    comment.refresh_from_db()
+    upvote_comment.refresh_from_db()
+    assert response.status_code == 200
+    assert comment in response.context["posts"]
```
jacoor:
```suggestion

```
This test is the same as the `test_filter_all_post`  as w/o filtering it returns all data. Please update `test_filter_all_post`

jacoor (changes requested):

jacoor (approved):

PR #131, opened 2025-02-23
Fix comment editing, add urls for communities

konrad-paszek:
Fixes according to https://app.clickup.com/t/8697w013u & https://app.clickup.com/t/8697w00vy

## Summary by Sourcery

This pull request introduces comment editing functionality and enhances community links throughout the application. It enables users to modify their existing comments and provides direct navigation to community details from various post listings.

New Features:
- Adds the ability for users to edit their comments.

Enhancements:
- Updates the post header and user post list templates to link to community detail pages.

### DIFF
--- a/core/tests/test_views.py
+++ b/core/tests/test_views.py
@@ -692,7 +692,6 @@ def test_edit_comment_valid(client: Client, user: User, post: Post, comment: Pos
     client.force_login(user)
     edit_response = client.post(reverse("edit_comment", kwargs={"pk": comment.pk}), data={"content": "TEST"})
     assert edit_response.status_code == 302
-    post.refresh_from_db()
     response = client.get(reverse("post-detail", kwargs={"pk": comment.pk}))
     assert response.context["post"].content == "TEST"
 
--- a/core/tests/test_views.py
+++ b/core/tests/test_views.py
@@ -127,6 +127,13 @@ def restricted_community(user: User) -> Community:
     return Community.objects.create(name="Restricted Community", is_active=True, author=user)
 
 
+@pytest.fixture()
+def comment(post: Post, community: Community, user: User) -> Post:
+    return Post.objects.create(
+        parent_id=post.pk, community=community, author=user, content="This is a test comment content."
+    )
+
+
 def test_add_post_valid(client: Client, user: User, community: Community) -> None:
     data = {
         "community": community.pk,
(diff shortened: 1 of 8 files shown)


core/tests/test_views.py
```diff
@@ -679,3 +679,22 @@ def test_fetch_user_comments_with_empty_filter(client: Client, user: User, comme
     response = client.get(f"{reverse('user_comments')}?filter=")
     assert response.status_code == 200
     assert len(response.context_data["comments"]) == 1
+
+
+def test_edit_comment_valid(client: Client, user: User, post: Post) -> None:
+    data = {
```
jacoor: Thank you for this test, this is very good excercise for test optimisation. 

Hmm this test is kinda weird. Let me describe what I see it does
1. It creates a comment with data from line 685
2. Verifies the comment was created (200) but does not prove the data in the post is what we send, or the parent is correct. 
3. edits the comments
4. Verifies it's ok by checking 302 http code, which is not 200 OK - error page could also be moved. 
5. Refreshes post from DB but does not to anything with that post.
6. Verifies the content after edition is TEST.

THe name of the test is`test_edit_comment_valid` and it creates comment, checks comment creation and later edition.  It does much more than expected. 

Please edit the test to only check editing. Use fixture for post and comment, if not already created. 
The test should follow structure
Given, when, then. The current version is like: 
Given - user. data.
when -> create comment
then -> comment created
When(?) edit response
Then.... 

what I would do is
- Given: use fixture for post and first comment
- When: edit the comment by HTTP request
- Then: verify the comment by retriebving it from database - the call in 699 to post details is not needed. You already have this data in DB and in edit response context. 
The provided steps will make test much faster by using only one HTTP call instead of 3.

jacoor (changes requested):
Very good. The test need optimisation.

core/tests/test_views.py
```diff
@@ -679,3 +679,22 @@ def test_fetch_user_comments_with_empty_filter(client: Client, user: User, comme
     response = client.get(f"{reverse('user_comments')}?filter=")
     assert response.status_code == 200
     assert len(response.context_data["comments"]) == 1
+
+
+def test_edit_comment_valid(client: Client, user: User, post: Post) -> None:
```
jacoor: you could also add one test for trying to save empty content.

core/tests/test_views.py
```diff
@@ -679,3 +686,19 @@ def test_fetch_user_comments_with_empty_filter(client: Client, user: User, comme
     response = client.get(f"{reverse('user_comments')}?filter=")
     assert response.status_code == 200
     assert len(response.context_data["comments"]) == 1
+
+
+def test_edit_comment_valid(client: Client, user: User, post: Post, comment: Post) -> None:
+    client.force_login(user)
+    edit_response = client.post(reverse("edit_comment", kwargs={"pk": comment.pk}), data={"content": "TEST"})
+    assert edit_response.status_code == 302
+    post.refresh_from_db()
```
jacoor: Not needed. You are checking HTTP response not DB response. This is not used.

jacoor (approved):

jacoor (changes requested):
This close

jacoor (approved):

PR #140, opened 2025-05-10
Add sending email when new reported post/comment is created

konrad-paszek:
Adding changes according to this tickets -> https://app.clickup.com/t/8698n74kr & https://app.clickup.com/t/8698n7547

## Summary by Sourcery

Add email notification for moderators when a new post or comment is reported

New Features:
- Implement email notification to superusers when a post is reported

Enhancements:
- Add tooltips to improve user guidance in moderator dashboard
- Include a help section in the moderator dashboard

Tests:
- Update existing tests to verify email notification functionality

### DIFF
--- a/core/templates/core/moderator_dashboard.html
+++ b/core/templates/core/moderator_dashboard.html
@@ -4,6 +4,7 @@
 
 {% block extra_css %}
 <link rel="stylesheet" href="{% static 'core/css/style.css' %}">
+<link rel="stylesheet" href="{% static 'core/css/help.css' %}">
 {% endblock %}
 
 {% block content %}
@@ -17,7 +18,7 @@
 
     <div class="row">
         <div class="col-md-4">
-            <div class="card text-white bg-primary mb-3">
+            <div class="card text-white bg-primary mb-3" data-tooltip="Shows the total number of posts on the platform.">
                 <div class="card-header">Total Posts</div>
                 <div class="card-body">
                     <h5 class="card-title">{{ active_posts }}</h5>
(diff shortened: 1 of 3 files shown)


core/templates/core/moderator_dashboard.html
jacoor, line 1: @konrad-paszek please explain changes in this file. The pull request is about sending email notification, not HTML content changes. 
Please either revert or go over the file and add comments about what are those changes.

core/views.py
```diff
@@ -440,6 +442,16 @@ def form_valid(self: "PostReportView", form: forms.ModelForm) -> HttpResponseRed
         post_report.post = post
         post_report.report_person = report_person
         post_report.save()
+        full_activation_link = self.request.build_absolute_uri(reverse("reported-post", kwargs={"pk": post.id}))
+        mail_status = send_mail(
+            "New reported post",
+            f"A new post has been reported and is awaiting your review here: {full_activation_link}.",
+            settings.DEFAULT_FROM_EMAIL,
+            User.objects.filter(is_superuser=True).values_list("email", flat=True),
+            fail_silently=True,
```
jacoor: I would say fail silently false. We need to see the error, or have some kind of report. RIght now only the user will see the error, no one else...
For example sentry will not catch it (if implemented) because there is no exception raised. 
It also breaks Zen of Python:
`Errors should never pass silently.`
At least log is required.

jacoor (changes requested):

core/templates/core/moderator_dashboard.html
konrad-paszek: It is related to second ticket about adding help section, I see the formatting has changed not sure why but logic is with new help section

core/views.py
konrad-paszek: Hm, yes you are right but I don't know how to do it in correct way, we don't have logger, maybe we should discuss about it and create new epic/ticket

jacoor: just make it fail loud. It's better to fail loud than to hide.

jacoor (changes requested):

jacoor (approved):

